Issue #115, opened 2015-01-28
Extract major, minor, patch version


Michenux:
Hi,

Could you add some methods in the api to extract major/minor/patch ?

For example : 
sem.major('1.2.3') => 1
sem.minor('1.2.3') => 2
sem.patch('1.2.3') => 3
sem.trimMinor('1.2.3') => 1.2

Thanks

isaacs:
Sounds reasonable. Patch welcome.

Hypercubed:
I submitted a pull request for this feature... although it is really just sugar for `new SemVer('1.2.3').major;`, etc.

JamesMGreene:
@Hypercubed's PR #116 was released a while ago in [`v4.3.0`](https://github.com/npm/node-semver/releases/tag/v4.3.0). 

Seems like this issue should be closed.

isaacs:
Thanks!

PR #116, opened 2015-02-03
Added functions to extract major, minor, patch version

Hypercubed:
Added functions to extract major, minor, and patch version from string as suggested in #115.


isaacs:
Landed on master, but please use spaces for indentation next time.

PR #119, opened 2015-02-24
fix github URL

hoobdeebla:
The repository is now under the ownership of npm's organization, so package.json should reflect that change


isaacs:
Thanks!

PR #125, opened 2015-04-12
Make "*" not satisfied by prerelease versions

### LINKED ISSUE #123: "*" is satisfied by versions with a pre-release
In the range tests there's this test:
```
['*', 'v1.2.3-foo', true]
```

Which indicates that the version `1.2.3-foo` will satisfy the range `*`. But this behaviour goes against what is described in the README, which says that:

> If a version has a prerelease tag (for example, 1.2.3-alpha.3) then it will only be allowed to satisfy comparator sets if at least one comparator with the same [major, minor, patch] tuple also has a prerelease tag.

I can't find anywhere that states that X-ranges allow pre-release versions. I would expect that `1.2.3-foo` would not satisfy `*`, but if this is the expected behaviour, then it should be documented.

Also, if `*` is satisfied by `1.2.3-foo`, then why is `1.*` not satisfied by `1.2.3-foo`? This behaviour seems inconsistent.

adamreeve:
Fixes #123 

In the process of fixing this, I found that `ltr('>=0.0.0', '1.2.3-foo') === true`. I'm not sure how to sensibly handle this case, but I documented the behaviour with a new test.

### DIFF
--- a/semver.js
+++ b/semver.js
@@ -1170,6 +1170,9 @@ function outside(version, range, hilo, loose) {
     var low = null;
 
     comparators.forEach(function(comparator) {
+      if (comparator.semver === ANY) {
+        comparator = new Comparator('>=0.0.0')
+      }
       high = high || comparator;
       low = low || comparator;
       if (gtfn(comparator.semver, high.semver, loose)) {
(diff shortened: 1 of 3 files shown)


adamreeve:
Closing as these were committed to master. See:
* 11cbbb1fb37155aff2cadc098e83d606da1736c5
* 93a870665bba0f4be5c4edc19d6f94ff39ad60e8
* aad3543b8c3f767f80ad3a194b281f900b488218

PR #136, opened 2015-08-30
remove unused uglify-js

stevemao:


isaacs:
Landed on fa88737 some time ago.  Thanks!

Issue #133, opened 2015-08-14
Add method to test if tag is a pre-release tag


SamyPesse:
It'd be great to have a method like `semver.isPrerelease(v)`:

```js
> semver.isPrerelease("1.0.0")
false

> semver.isPrerelease("1.0.0-alpha.1")
true
```

SamyPesse:
And maybe have a method to extract the type of prerelease:

```js
> semver.prerelease("1.0.0-alpha.1") // or semver.pre ?
"alpha"

> semver.prerelease("1.0.0-beta.45")
"beta"
```

isaacs:
Shouldn't it return the full `"alpha.1"` or `"beta.45"`?

Another option would be to return the full parsed `prerelease` field if it's there, or else `null`.  Maybe something like this?

```
> require('semver').parse('1.2.3-alpha.23.bravo.43').prerelease
[ 'alpha', 23, 'bravo', 43 ]
> require('semver').parse('1.2.3').prerelease
[]
> function prerelease (ver) { var p = require('semver').parse(ver); return p && p.prerelease.length && p.prerelease || null }
undefined
> prerelease('1.2.3-alpha.4.bravo.65')
[ 'alpha', 4, 'bravo', 65 ]
> prerelease('1.2.3')
null
```

SamyPesse:
Something like this would be perfect :+1:

You're right, it should return the full parsed `prerelease`, since there is no strict convention on the prerelease's format: some people are using `alpha.x` and some others are using numerical values like `1.0.0-0.3.7`; and others are using the pre-release part as a patch (ex: https://github.com/nwjs/npm-installer#install-a-specific-version-of-nwjs).

I was trying to find the most conventional way to filter versions by "pre-release's category" (`stable`, `beta`, `alpha`, ..) for [Nuts](https://github.com/GitbookIO/nuts). I'll probably consider that pre-release like `1.0.0-0.3.7` are `alpha`.

luckydrq:
I have the same demand. Any update on this?

PR #144, opened 2015-12-07
Add prerelease function

ekmartin:
Implements the functionality discussed in #133.
`prerelease('1.2.3-alpha.1') -> ['alpha', 1]`
`prerelease('1.2.3') -> null`

### DIFF
--- a/semver.js
+++ b/semver.js
@@ -1186,3 +1186,9 @@ function outside(version, range, hilo, loose) {
   }
   return true;
 }
+
+exports.prerelease = prerelease;
+function prerelease(version, loose) {
+  var parsed = parse(version, loose);
+  return (parsed && parsed.prerelease.length) ? parsed.prerelease : null;
+}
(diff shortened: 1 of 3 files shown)


semver.js
```diff
@@ -1186,3 +1186,9 @@ function outside(version, range, hilo, loose) {
   }
   return true;
 }
+
+exports.prerelease = prerelease;
+function prerelease(version) {
```
isaacs: should take a second "loose" boolean arg, and pass that to the new Semver in the next line.

ekmartin: Makes sense, should have seen that. Added now, and updated the test assertions.

JPeColCom:
happens something here?

ljharb:
Currently I'm using `version.indexOf('-') > -1` to determine `isPrerelease` but I'd really love to use a more robust/centralized method of doing this. Any progress here?

isaacs:
Landed and published on 5.2.  Thanks!

PR #145, opened 2015-12-07
Add Node 4 and 5 to .travis.yml

ekmartin:


isaacs:
Thanks for the nudge.  I updated it to match all my other projects.

PR #151, opened 2016-02-02
Add a prerelease getter

dbyzero:
We have `major`, `minor`, `patch` getter... Let's the `prerelease` getter coming !


dbyzero:
Dupplicated by PR #144

PR #152, opened 2016-02-11
Fix Backus-Naur "nr" expansion (previous expansion forbade 1 to 9)

ixalon:
The old BNF expands "nr"s as:
- 0, or
- 1-9 followed by **one** or more digit from 0-9

i.e. excludes non-zero single-digit numbers. It should be:
- 0, or
- 1-9 followed by **zero** or more digit from 0-9


isaacs:
This is correct.

Issue #149, opened 2016-01-12
Allow piping into semver to read version


felixfbecker:
It would be very nice if we could do this:
```sh
node -v | semver -r ^5.4
```

isaacs:
I'm ok with it.  Patch welcome, as long as it includes docs and test :)

felixfbecker:
Great. Is it possible to "detect" if there is data being piped from stdin or is a new command line flag needed?

isaacs:
You can check process.stdin.isTTY and also default to stdin if - is an option or if no options are provided. The stdin stream should be a white space delimited set of versions, probably?

felixfbecker:
> You can check process.stdin.isTTY

thanks.
> The stdin stream should be a white space delimited set of versions, probably?

I agree.

There is no engine field in package.json. What node versions have to be supported? I guess no arrow functions?

felixfbecker:
I guess the whitespace splitting will not work completely because one version can contain spaces, like `>=0.2 <=0.8`. The only option is to split by line breaks.

isaacs:
@felixfbecker Those are version ranges, not versions.

I'm imagining a thing like this:

```bash
get-list-of-versions-somehow | semver -r '>=2' > file-with-versions-gt-two.txt
```

So, the list of *versions* can be piped to stdin, but not the set of *ranges*

gWOLF3:
can we merge this?

PR #161, opened 2016-06-04
Allow piping into semver to read versions

richardgarnier:
This is a patch for #149.

Before entering the main, we check if the process is piped. If so then we fill the `versions` array and then we launch the main. In order to fill the array, we split on spaces.

NB : versions can be accepted both from the pipe and as arguments.
For example `echo 1.0.0 | semver -r '>=1' 1.1.1` is accepted.

NB2 : testing relies on comparing the output of the version without piping, and the version with piping.

### DIFF
--- a/bin/semver
+++ b/bin/semver
@@ -16,10 +16,29 @@ var argv = process.argv.slice(2)
   , semver = require("../semver")
   , reverse = false
 
-main()
+preMain()
 
-function main () {
-  if (!argv.length) return help()
+function preMain() {
+  if (!process.stdin.isTTY) {
+    process.stdin.setEncoding('utf8');
+
+    process.stdin.on('readable', function() {
+      var chunk = process.stdin.read();
+      if (chunk !== null) {
+        versions = versions.concat(chunk.split(/\s/));
+      }
+    });
+
+    process.stdin.on('end', () => {
+      main(versions.length !== 0);
+    });
+  } else {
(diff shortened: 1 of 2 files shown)


bin/semver
```diff
@@ -56,6 +75,7 @@ function main () {
         break
       case "-h": case "--help": case "-?":
         return help()
+        break
```
akatsukle: I guess the break here was missing. This is not the main purpose of the patch though.

akatsukle:
Well that test was embarrassing... sorry

bin/semver
isaacs: A break isn't strictly necessary here, since the `return` is just as effective at ending the block.  But if a linter or something is complaining about it, whatevdr.

isaacs:
This is ok, and seems to work, but it seems like it'd be better to open stdin, and parse/filter each version as lines are read.  That way it could be done interactively.  I could see this being really handy if you want to test out whether a given range works with various versions.

You can use node's builtin `readline` module to do this.

isaacs:
Oh, I just realized something about this.

This means that any programs that `spawn('semver')` are going to hang until stdin is closed.

That's probably surprising.  One approach would be to only read the stdin pipe if `-` is passed as an argument.

wraithgar:
Closing this one for now since it still has work to be done.

PR #163, opened 2016-07-13
add minSatisfying

wraithgar:
This adds the function `minSatisfying` which is a mirror of `maxSatisfying` i.e. it returns the *lowest* version in the range that satisfies the given range.  As you can see it's an exact copy of `minSatisfying` that simply reverses the sort order being passed to `.filter`

The need for this arose from trying to resolve multiple security advisories that have semver ranges for which the problem has been fixed.  It is valuable to be able to distill the minimum version needed to satisfy each of those ranges, to (as much as possible) mitigate the potential impact that updates would cause.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,6 +327,8 @@ strings that they parse.
   range.
 * `maxSatisfying(versions, range)`: Return the highest version in the list
   that satisfies the range, or `null` if none of them do.
+* `minSatisfying(versions, range)`: Return the lowest version in the list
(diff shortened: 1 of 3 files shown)


isaacs:
Good use case, well executed patch, thanks very much!

Landed and published on 5.3.0.

ashleygwilliams:
This is so cool! A lot of people have been asking for this. Thanks so much @wraithgar !!

PR #172, opened 2016-10-08
Added travis ci support for node 6

amilajack:


isaacs:
Thanks for the nudge.  Opted to keep the yaml formatting unchanged.

PR #173, opened 2016-10-08
Added syntax highlighting to readme example

amilajack:


isaacs:
Landed on aaba7103e1837c4ecbfe8aff956d08adf67240a1 with slight modification for some more picky markdown parsers.

PR #179, opened 2016-12-11
minimize heap allocations for maxSatisfying/minSatisfying

zkat:
It's a little bit of code golf, but I think it's readable enough, and can have some significant improvements depending on the size of arrays being compared. If anyone ever actually puts this in hot code. :)


isaacs:
Landed on 32802c53503f67e5ebfd8c59aadf0fcec5d6a4a1

PR #184, opened 2017-01-27
Adding Keywords

evantsyferov:
I have added some keywords to the package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,11 @@
 {
   "name": "semver",
   "version": "5.3.0",
+  "keywords":{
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,11 @@
 {
   "name": "semver",
   "version": "5.3.0",
+  "keywords":{
```
humphd: Becuase this is a list of keywords, you want to use `[` vs. `{`, so it should look like:

```js
"keywords": [
...
],
```

PR #187, opened 2017-04-01
Add support for range and comparators comparison

### LINKED ISSUE #178: Feature request: compare two ranges
I'm trying to compare an existing semver range in a package.json, with a peerDependency range in another package.json - in other words, I want to see if the two ranges are compatible (ie, if there exists a version that satisfies both).

I tried playing with `Range` but couldn't come up with anything.

A simple example would be, `^1.2.3` and `<= 1.3.0` should return either a semver range that satisfies both, or the highest version number that satisfies both (either is sufficient).

Is this possible?

rtfpessoa:
* Check if two ranges intersect
* Check if one comparator satisfies a range
* Check if two comparators intersect

Fixes #178

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .*.swp
 coverage/
 .nyc_output/
+.idea
--- a/README.md
+++ b/README.md
@@ -329,6 +329,10 @@ strings that they parse.
   (`major`, `premajor`, `minor`, `preminor`, `patch`, `prepatch`, or `prerelease`),
   or null if the versions are the same.
 
+### Comparators
+* `intersects(comparator)`: Return true if the comparators intersect
+* `satisfiesRange(range)`: Return true if the comparator intersects with any
+of the range's comparators
 
 ### Ranges
 
@@ -347,6 +351,7 @@ strings that they parse.
   the bounds of the range in either the high or low direction.  The
   `hilo` argument must be either the string `'>'` or `'<'`.  (This is
   the function called by `gtr` and `ltr`.)
+* `intersects(range)`: Return true if any of the ranges comparators intersect
 
 Note that, since ranges may be non-contiguous, a version might not be
 greater than a range, less than a range, *or* satisfy a range!  For
--- a/README.md
+++ b/README.md
@@ -331,8 +331,6 @@ strings that they parse.
 
 ### Comparators
 * `intersects(comparator)`: Return true if the comparators intersect
-* `satisfiesRange(range)`: Return true if the comparator intersects with any
-of the range's comparators
 
 ### Ranges
 
--- a/semver.js
+++ b/semver.js
@@ -704,6 +704,48 @@ Comparator.prototype.test = function(version) {
   return cmp(version, this.operator, this.semver, this.loose);
 };
 
+Comparator.prototype.intersects = function(comp, loose, platform) {
+  if (!(comp instanceof Comparator)) {
+    throw new TypeError('a Comparator is required');
+  }
+
+  var rangeTmp;
+
+  if (this.operator === '') {
+    rangeTmp = new Range(comp.value, loose, platform);
+    return satisfies(this.value, rangeTmp, loose, platform);
+  } else if (comp.operator === '') {
+    rangeTmp = new Range(this.value, loose, platform);
+    return satisfies(comp.semver, rangeTmp, loose, platform);
+  }
+
+  var sameDirectionIncreasing = (this.operator === '>=' || this.operator === '>') && (comp.operator === '>=' || comp.operator === '>');
+  var sameDirectionDecreasing = (this.operator === '<=' || this.operator === '<') && (comp.operator === '<=' || comp.operator === '<');
+  var sameSemVer = this.semver.raw === comp.semver.raw;
+  var differentDirectionsInclusive = (this.operator === '>=' || this.operator === '<=') && (comp.operator === '>=' || comp.operator === '<=');
+  var oppositeDirectionsLessThan = cmp(this.semver, '<', comp.semver, loose) &&
+    ((this.operator === '>=' || this.operator === '>') && (comp.operator === '<=' || comp.operator === '<'));
+  var oppositeDirectionsGreaterThan = cmp(this.semver, '>', comp.semver, loose) &&
+    ((this.operator === '<=' || this.operator === '<') && (comp.operator === '>=' || comp.operator === '>'));
+
+  return sameDirectionIncreasing || sameDirectionDecreasing || (sameSemVer && differentDirectionsInclusive) ||
+    oppositeDirectionsLessThan || oppositeDirectionsGreaterThan;
+};
+
+Comparator.prototype.satisfiesRange = function(range, loose, platform) {
+  if (!(range instanceof Range)) {
+    throw new TypeError('a Range is required');
+  }
+
+  var comp = this;
+
+  return range.set.some(function(comparators) {
+    return comparators.every(function(comparator) {
+      return comp.intersects(comparator, loose, platform);
+    });
+  });
+};
+
 
 exports.Range = Range;
 function Range(range, loose) {
@@ -783,6 +825,18 @@ Range.prototype.parseRange = function(range) {
   return set;
 };
 
+Range.prototype.intersects = function(range, loose, platform) {
+  if (!(range instanceof Range)) {
+    throw new TypeError('a Range is required');
+  }
+
+  return this.set.some(function(comparators) {
+    return comparators.every(function(comparator) {
+      return comparator.satisfiesRange(range, loose, platform);
+    });
+  });
+};
+
 // Mostly just for testing and legacy API reasons
 exports.toComparators = toComparators;
 function toComparators(range, loose) {
@@ -790,56 +844,6 @@ function toComparators(range, loose) {
     return comp.map(function(c) {
       return c.value;
     }).join(' ').trim().split(' ');
-  });
-}
-
-exports.comparatorsIntersect = comparatorsIntersect;
-function comparatorsIntersect(compA, compB, loose, platform) {
-  compA = new Comparator(compA, loose);
-  compB = new Comparator(compB, loose);
-
-  if (compA.operator === '') {
-    var rangeB = new Range(compB.value, loose, platform);
-    return satisfies(compA.value, rangeB, loose, platform);
-  } else if (compB.operator === '') {
-    var rangeA = new Range(compA.value, loose, platform);
-    return satisfies(compB.semver, rangeA, loose, platform);
-  }
-  // Same direction increasing
-  return ((compA.operator === '>=' || compA.operator === '>') && (compB.operator === '>=' || compB.operator === '>')) ||
-    // Same direction decreasing
-    ((compA.operator === '<=' || compA.operator === '<') && (compB.operator === '<=' || compB.operator === '<')) ||
-    // Different directions, same semver and inclusive operator
-    (compA.semver.raw === compB.semver.raw &&
-    (compA.operator === '>=' || compA.operator === '<=') && (compB.operator === '>=' || compB.operator === '<=')) ||
-    // Opposite matching directions
-    (cmp(compA.semver, '<', compB.semver, loose) &&
-    ((compA.operator === '>=' || compA.operator === '>') && (compB.operator === '<=' || compB.operator === '<'))) ||
-    (cmp(compA.semver, '>', compB.semver, loose) &&
-    ((compA.operator === '<=' || compA.operator === '<') && (compB.operator === '>=' || compB.operator === '>')));
-}
-
-exports.comparatorSatisfiesRange = comparatorSatisfiesRange;
-function comparatorSatisfiesRange(comp, range, loose, platform) {
-  comp = new Comparator(comp, loose);
-  range = new Range(range, loose, platform);
-
-  return range.set.some(function(comparators) {
-    return comparators.every(function(comparator) {
-      return comparatorsIntersect(comparator, comp, loose, platform);
-    });
-  });
-}
-
-exports.rangesIntersect = rangesIntersect;
-function rangesIntersect(rangeA, rangeB, loose, platform) {
-  rangeA = new Range(rangeA, loose, platform);
-  rangeB = new Range(rangeB, loose, platform);
-
-  return rangeA.set.some(function(comparators) {
-    return comparators.every(function(comparator) {
-      return comparatorSatisfiesRange(comparator, rangeB, loose, platform);
-    });
   });
 }
 
--- a/semver.js
+++ b/semver.js
@@ -743,20 +743,6 @@ Comparator.prototype.intersects = function(comp, loose) {
     oppositeDirectionsLessThan || oppositeDirectionsGreaterThan;
 };
 
-Comparator.prototype.satisfiesRange = function(range, loose) {
-  if (!(range instanceof Range)) {
-    throw new TypeError('a Range is required');
-  }
-
-  var comp = this;
-
-  return range.set.some(function(comparators) {
-    return comparators.every(function(comparator) {
-      return comp.intersects(comparator, loose);
-    });
-  });
-};
-
 
 exports.Range = Range;
 function Range(range, loose) {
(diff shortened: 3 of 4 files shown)


semver.js
```diff
@@ -793,6 +793,56 @@ function toComparators(range, loose) {
...
+    (compA.semver.raw === compB.semver.raw &&
+    (compA.operator === '>=' || compA.operator === '<=') && (compB.operator === '>=' || compB.operator === '<=')) ||
+    // Opposite matching directions
+    (cmp(compA.semver, '<', compB.semver, loose) &&
+    ((compA.operator === '>=' || compA.operator === '>') && (compB.operator === '<=' || compB.operator === '<'))) ||
+    (cmp(compA.semver, '>', compB.semver, loose) &&
+    ((compA.operator === '<=' || compA.operator === '<') && (compB.operator === '>=' || compB.operator === '>')));
+}
+
+exports.comparatorSatisfiesRange = comparatorSatisfiesRange;
```
shellscape: function names are a little overly verbose/long for my tastes. but looks like good functionality

rtfpessoa:
@shellscape I am completely open for alternatives. Let me know if you have any suggestions.

semver.js
```diff
@@ -793,6 +793,56 @@ function toComparators(range, loose) {
   });
 }
 
+exports.comparatorsIntersect = comparatorsIntersect;
+function comparatorsIntersect(compA, compB, loose, platform) {
```
ljharb: Is this better as a standalone function, or as a prototype method on Comparator?

semver.js
```diff
@@ -793,6 +793,56 @@ function toComparators(range, loose) {
...
+  compA = new Comparator(compA, loose);
+  compB = new Comparator(compB, loose);
```
ljharb: Reassigning function arguments deopts in v8; could you use new variables for these?

semver.js
ljharb: Similarly; if this was a prototype method on Comparator it could just be `satisfiesRange`?

semver.js
```diff
@@ -793,6 +793,56 @@ function toComparators(range, loose) {
...
+
+  return range.set.some(function(comparators) {
+    return comparators.every(function(comparator) {
+      return comparatorsIntersect(comparator, comp, loose, platform);
+    });
+  });
+}
+
+exports.rangesIntersect = rangesIntersect;
+function rangesIntersect(rangeA, rangeB, loose, platform) {
```
ljharb: And if this was a prototype method on Range it could just be `intersects`?

test/index.js
```diff
@@ -712,3 +712,77 @@ test('\nmin satisfying', function(t) {
...
+  t.end();
+});
+
+test('\nranges intersect', function(t) {
+  [
+    ['1.3.0 || <1.0.0 >2.0.0', '1.3.0 || <1.0.0 >2.0.0', true],
+    ['<1.0.0 >2.0.0', '>0.0.0', true],
+    ['<1.0.0 >2.0.0', '>1.4.0 <1.6.0', false],
+    ['<1.0.0 >2.0.0', '>1.4.0 <1.6.0 || 2.0.0', false],
+    ['<1.0.0 >=2.0.0', '2.1.0', false],
```
ljharb: Could you add a test case for a single version returning true?

test/index.js
```diff
@@ -712,3 +712,77 @@ test('\nmin satisfying', function(t) {
...
+    ['<1.0.0 >=2.0.0', '>1.4.0 <1.6.0 || 2.0.0', false]
+  ].forEach(function(v) {
+    var range1 = v[0];
+    var range2 = v[1];
+    var expect = v[2];
+    var actual = semver.rangesIntersect(range1, range2);
+    t.equal(actual, expect);
```
ljharb: Could we also assert that if you flip the range order that it provides the same value?

test/index.js
```diff
@@ -712,3 +712,77 @@ test('\nmin satisfying', function(t) {
...
+    ['<1.2.0', '<1.3.0', true],
+    ['<=1.2.0', '<1.3.0', true],
+    ['<1.2.0', '<=1.3.0', true],
+    // Different directions, same semver and inclusive operator
+    ['>=1.3.0', '<=1.3.0', true],
+    ['>=1.3.0', '>=1.3.0', true],
+    ['<=1.3.0', '<=1.3.0', true],
+    ['>1.3.0', '<=1.3.0', false],
+    ['>=1.3.0', '<1.3.0', false],
+    // Opposite matching directions
```
ljharb: This could instead be done in the iterator function, for every test, no?

rtfpessoa: I am not understanding exactly what you mean here. Can you extend a bit?

Also, should I amend my previous commit or should I create new commits to fix the requested changes?

semver.js
rtfpessoa: Sure, I think I was following the `toComparators` logic but I think I like it more this way too.

semver.js
rtfpessoa: 👍

semver.js
rtfpessoa: 👍

semver.js
rtfpessoa: 👍

test/index.js
rtfpessoa: 👍

test/index.js
rtfpessoa: 👍

semver.js
```diff
@@ -704,6 +704,42 @@ Comparator.prototype.test = function(version) {
   return cmp(version, this.operator, this.semver, this.loose);
 };
 
+Comparator.prototype.intersects = function(compOrStr, loose, platform) {
+  var rangeTmp;
+  var comp = new Comparator(compOrStr, loose);
```
ljharb: Would it be bad if this only took a Comparator? That would make the API more explicit and delegate more decisions to the user.

semver.js
```diff
@@ -704,6 +704,42 @@ Comparator.prototype.test = function(version) {
...
+    // Opposite matching directions
+    (cmp(this.semver, '<', comp.semver, loose) &&
+    ((this.operator === '>=' || this.operator === '>') && (comp.operator === '<=' || comp.operator === '<'))) ||
+    (cmp(this.semver, '>', comp.semver, loose) &&
+    ((this.operator === '<=' || this.operator === '<') && (comp.operator === '>=' || comp.operator === '>')));
+};
+
+Comparator.prototype.satisfiesRange = function(rangeOrStr, loose, platform) {
+  var comp = this;
+  var range = new Range(rangeOrStr, loose, platform);
```
ljharb: Same question here

semver.js
```diff
@@ -783,6 +819,16 @@ Range.prototype.parseRange = function(range) {
   return set;
 };
 
+Range.prototype.intersects = function(rangeOrStr, loose, platform) {
+  var range = new Range(rangeOrStr, loose, platform);
```
ljharb: And here

test/index.js
ljharb: I mean, line 752 asks if comparator1 intersects comparator2; could you also add an assertion after that asking if comparator2 intersects comparator1?

ljharb: I think amending and force pushing is always fine; unless one of this repo's collabs says different.

semver.js
rtfpessoa: I am ok with this change, but I actually do not think the user makes more decisions. Since we allow the method to receive either a comparator or a string they can do all the same decisions if they want.

But the API is definitely more explicit if we force a Comparator.

Usually I am a bit more defensive and that is why I usually try to handle all the weird possibilities. But if this was Typescript it would make much more sense to receive only the comparator.

rtfpessoa:
@ljharb I think I made all the changes you suggested in this last commit. Take a look when you have some time.

semver.js
ljharb: I don't think typescript has anything to do with it; it's nicer to have monomorphic APIs with or without types. By "decisions" I mean, the comparator and the range having the same `loose` setting, for example - forcing the user to make that choice means they can never be ignorant of the default.

semver.js
```diff
@@ -704,6 +704,40 @@ Comparator.prototype.test = function(version) {
...
+
+  if (this.operator === '') {
+    rangeTmp = new Range(comp.value, loose, platform);
+    return satisfies(this.value, rangeTmp, loose, platform);
+  } else if (comp.operator === '') {
+    rangeTmp = new Range(this.value, loose, platform);
+    return satisfies(comp.semver, rangeTmp, loose, platform);
+  }
+  // Same direction increasing
+  return ((this.operator === '>=' || this.operator === '>') && (comp.operator === '>=' || comp.operator === '>')) ||
```
ljharb: I feel like making this one statement makes it a bit more confusing to read; what do you think about using if/else here?

rtfpessoa: Can you provide an example?

rtfpessoa:
When you say "I would strongly prefer much more defensive throwing to check the types of arguments at runtime, but il leave insisting on that to the collaborators", like what check if `instanceof`?

semver.js
ljharb: Sure - maybe just separate variables would help. Here's a subset:
```js
var sameDirectionIncreasing = (this.operator === '>=' || this.operator === '>') && (comp.operator === '>=' || comp.operator === '>');
var sameDirectionDecreasing = (this.operator === '<=' || this.operator === '<') && (comp.operator === '<=' || comp.operator === '<');
var sameSemver = this.semver.raw === comp.semver.raw;
var differentDirectionsInclusive = (this.operator === '>=' || this.operator === '<=') && (comp.operator === '>=' || comp.operator === '<=');
return sameDirectionIncreasing || sameDirectionDecreasing || (sameSemver && differentDirectionsInclusive);
```

something like that? the long chain of boolean operations makes it hard for me to read.

ljharb:
@rtfpessoa exactly like that - i'd do `if (!(range instanceof Range)) { throw new TypeError('a Range is required'); }`, `if (typeof loose !== 'undefined' || typeof loose !== 'boolean') { throw new TypeError('loose must be a boolean if present'); }` and things like that.

In other words, if I'm passing the wrong kind of argument, I want to be alerted to that fact as soon and explicitly as possible.

semver.js
rtfpessoa: Sounds like a nice improvement.

rtfpessoa:
Just did both changes except for the `loose`  and `platform` check since I never use them and for that reason it would make more sense to be checked in `Range` or `Comparator`.

rtfpessoa:
Also, I added a couple more tests for the parameter checks.

rtfpessoa:
Also, I added a couple more tests for the parameter checks and fixed the format issues.

iarna:
This seems super useful! It just needs documentation to land.

ljharb:
Awesome! If @rrfpesoa can't find time to write it soon, I'll try to find time to write some :-)

rtfpessoa:
@iarna can you extend a bit on what do you mean by documentation?
I could not find any docs explaining a proper contribution, maybe I looked in the wrong places. Can you point me in the right direction?

.gitignore
```diff
@@ -2,3 +2,4 @@
 .*.swp
 coverage/
 .nyc_output/
+.idea
```
ljharb: Generally this belongs in a global git config, rather than polluting specific git repos with IDE-specific knowledge (nbd tho)

iarna:
@rtfpessoa Yes, that's exactly what I was looking for!

rtfpessoa:
@iarna, not sure if you noticed but I already did it. Le me know if you need any other changes for the  merge.

rtfpessoa:
any updates about this @isaacs @iarna ?

franher:
looking forward for the landing of this PR 🙏

agebhardt:
Really excited to be able to use this feature! Are there any updates on this PR?

semver.js
```diff
@@ -704,6 +704,48 @@ Comparator.prototype.test = function(version) {
...
+
+  if (this.operator === '') {
+    rangeTmp = new Range(comp.value, loose, platform);
+    return satisfies(this.value, rangeTmp, loose, platform);
+  } else if (comp.operator === '') {
+    rangeTmp = new Range(this.value, loose, platform);
+    return satisfies(comp.semver, rangeTmp, loose, platform);
+  }
+
+  var sameDirectionIncreasing = (this.operator === '>=' || this.operator === '>') && (comp.operator === '>=' || comp.operator === '>');
```
isaacs: These really long lines are very hard to read.

Can you please break this up into separate stanzas?  There should be no lines longer than 80 chars under any circumstances.  70 is an even better limit.

semver.js
```diff
@@ -704,6 +704,48 @@ Comparator.prototype.test = function(version) {
...
+  var oppositeDirectionsLessThan = cmp(this.semver, '<', comp.semver, loose) &&
+    ((this.operator === '>=' || this.operator === '>') && (comp.operator === '<=' || comp.operator === '<'));
+  var oppositeDirectionsGreaterThan = cmp(this.semver, '>', comp.semver, loose) &&
+    ((this.operator === '<=' || this.operator === '<') && (comp.operator === '>=' || comp.operator === '>'));
+
+  return sameDirectionIncreasing || sameDirectionDecreasing || (sameSemVer && differentDirectionsInclusive) ||
+    oppositeDirectionsLessThan || oppositeDirectionsGreaterThan;
+};
+
+Comparator.prototype.satisfiesRange = function(range, loose, platform) {
```
isaacs: What is the "platform" variable being passed around here?  It doesn't seem like it's ever used for anything.

README.md
```diff
@@ -329,6 +329,10 @@ strings that they parse.
   (`major`, `premajor`, `minor`, `preminor`, `patch`, `prepatch`, or `prerelease`),
   or null if the versions are the same.
 
+### Comparators
+* `intersects(comparator)`: Return true if the comparators intersect
+* `satisfiesRange(range)`: Return true if the comparator intersects with any
```
isaacs: The naming of this is a bit weird.  You wouldn't normally say that a comparator "satisfies" a range.  A range is satisfied by a single _thing_ which has a specific version, not by a comparator, which is like a mini-range.  Also, a comparator like `<=1.4.5` would return true for this method, given the Range `>=1.3.0`, even though many versions allowed by that comparator are not satisfying answers for the Range.  We can only talk about intersections and unions with ranges (and their building blocks, Comparators) because they are sets, not discrete atomic items.

Having a `comparator.intersects(comparator)` method seems like a building block towards having range intersection.  People generally don't interact with comparators directly.  Having (effectively) `comparator.intersects(range)` is a bit odd, and using the language of "satisfying" makes it odder.

It seems like what you'd really want in most cases is `range.intersects(range)`.  It's probably better to leave aside things that don't get us closer to that.

isaacs: I think this is probably addressed by calling it `comparator.intersectsRange()`?  Or even cleaner, make `Comparator.intersects` take either a Range or Comparator as an argument.

semver.js
isaacs: For more context on this, I've got somewhat bad vision, and usually keep my font size turned way up. It's a pretty common accessibility issue, though.
 
image

rtfpessoa:
@isaacs will try to do this soon.

semver.js
rtfpessoa: Ported this from the Ruby version and somehow did not notice platform did not exist here.

rtfpessoa:
@isaacs let me know if this was what you asked. I removed the confusing method and exposed only the two intersections between the same type.

ljharb:
@isaacs @iarna @zkat any update on this one? I've been waiting on its release to fix an eslint bug for about 9 months now :-/

semver.js
```diff
@@ -704,6 +704,45 @@ Comparator.prototype.test = function(version) {
...
+    return satisfies(comp.semver, rangeTmp, loose);
+  }
+
+  var sameDirectionIncreasing =
+    (this.operator === '>=' || this.operator === '>') &&
+    (comp.operator === '>=' || comp.operator === '>');
+  var sameDirectionDecreasing =
+    (this.operator === '<=' || this.operator === '<') &&
+    (comp.operator === '<=' || comp.operator === '<');
+  var sameSemVer = this.semver.raw === comp.semver.raw;
```
isaacs: I think what you're after isn't `semver.raw` here, but `semver.version`.  Consider this:

```
> r2 = new s.Range('>=v1.2.3')
Range {
  loose: undefined,
  raw: '>=v1.2.3',
  set: [ [ [Object] ] ],
  range: '>=1.2.3' }
> r = new s.Range('<=1.2.3')
Range {
  loose: undefined,
  raw: '<=1.2.3',
  set: [ [ [Object] ] ],
  range: '<=1.2.3' }
> r.intersects(r2)
false
```

isaacs: Yep, that fixes it.  I'll include it in the merge, no worries.

isaacs:
Landed on master and published as v5.4.0.  Also added a top-level `semver.intersects(r1, r2)` command, and a bunch of tests to improve coverage.

ljharb:
Yay, thanks!

PR #188, opened 2017-04-01
Add support for string representations of numbers

Qard:
Pure numbers are error prone and hard to read. With string representations, one can be much more explicit about their versioning.

With a very terse format like `1.2.3` the probability of typos occurring, causing skipped numbers, increases. I propose a system-wide switch to string representations of numbers where versions would be written like `one dot two dot three` or `fourteen point five point three hundred fourty four`. I believe this format is much more explicit and error resilient than the purely numeric representation.

(April Fool's y'all! 😆 )

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,5 +18,8 @@
     "bin",
     "range.bnf",
     "semver.js"
-  ]
+  ],
+  "dependencies": {
+    "number-strings": "^1.0.1"
+  }
 }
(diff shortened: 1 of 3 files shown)


JelaniThompson:
Beautiful.

srl295:
Intl support? `tres punt tres punt dos` etc?

Qard:
Closing now, as April Fool's is over. I hope everyone got a laugh out of this. 😛

Issue #130, opened 2015-07-13
semver.satisfies(): '*' doesn't match prerelease identifiers 


myndzi:
`semver.satisfies('0.3.16-foo.1', '*')` -> `false`

megawac:
See https://github.com/npm/node-semver/issues/123

myndzi:
Even after browsing that issue, it is unclear to me why a specifier that literally means "anything" should not match... well, anything. Additionally, it appears that there is no way, then, to construct such a comparison?

isaacs:
@myndzi The principle at work here is that `0.3.16-foo.1` is a "prerelease" version.  Prereleases can include breaking changes at any time, and frequently don't even satisfy the API contracts inherent in a module, so they're excluded from any ranges that don't explicitly opt-in to a pre-release version in that chain.  So, for example, it *would* match `~0.3.16-alpha`, but that range would *not* be satisfied by `0.3.17-foo.1`, because that is a prerelease of a different version.

myndzi:
I can understand this, to an extent, in the context of a partial version (say, `1.*`), but `*` is literally "everything", so how can anything fall "outside" this range?

myndzi:
More importantly, I guess, it seems that the ability to opt in to or out of prerelease matches should more properly be an option passed to the `semver.satisfies` function, so that the user can control the behavior, rather than a forced behavior that only satisfies one of two needs...

isaacs:
Having an option to allow prerelease matches against non-prerelease version ranges would be fine.  It'd be tricky to get the API right, though, especially since there's already one boolean and adding another one edges a bit too far down the slippery slope of long function signatures.  Maybe it'd be ok to replace the `loose` params with an options object, or have `loose` matches also allow prerelease versions to match against ranges?

qzb:
In my opinion new x-ranges for prerelease versions would be a good idea. For example ``1.3.*-*`` should match to all prereleases  of ``1.3.*``, and ``* || *-*`` to any version at all (prerelease or not).

PR #192, opened 2017-04-18
Optional prerelease version lock

protyposis:
Prerelease versions are currently only satisfied by ranges containing a prerelease tag with the same `[major, minor, patch]` tuple as [documented](https://github.com/protyposis/node-semver/blob/aaba7103e1837c4ecbfe8aff956d08adf67240a1/README.md#prerelease-tags), which makes sense for the default use-cases of this library.

I currently have a use-case where I want to to specify version ranges of test cases to be run for certain versions of a tested software component. E.g. to run a test on all versions that ever exist, including prerelease versions (e.g. release candidates), I want to specify the range '*'. To run a test on all versions starting with prereleases of '2.0.0', I specify the range '>=2.0.0-0'. '>=2.0.0-0' should thus match '2.0.0-rc1', '2.0.0', '2.1.0-rc1', '2.2.0', etc., which is currently prevented by the `[major, minor, patch]` prerelease version lock.

To be able to use this library in such cases, I added an additional parameter **`prereleaseLock`** to the `satisfies` method (`satisfies(version, range, loose, prereleaseLock)`), which is `true` by default to get the default version locking behavior, and disables the version lock behavior when set to `false`. This is basically what has been described by @isaacs in #130 ([here](https://github.com/npm/node-semver/issues/130#issuecomment-121095949).

### DIFF
--- a/semver.js
+++ b/semver.js
@@ -1028,7 +1028,7 @@ function hyphenReplace($0,
 
 
 // if ANY of the sets match ALL of its comparators, then pass
-Range.prototype.test = function(version) {
+Range.prototype.test = function(version, prereleaseLock) {
   if (!version)
     return false;
 
@@ -1036,19 +1036,23 @@ Range.prototype.test = function(version) {
     version = new SemVer(version, this.loose);
 
   for (var i = 0; i < this.set.length; i++) {
-    if (testSet(this.set[i], version))
+    if (testSet(this.set[i], version, prereleaseLock))
       return true;
   }
   return false;
 };
 
-function testSet(set, version) {
+function testSet(set, version, prereleaseLock) {
   for (var i = 0; i < set.length; i++) {
     if (!set[i].test(version))
       return false;
   }
 
-  if (version.prerelease.length) {
+  if (prereleaseLock === undefined) {
+    prereleaseLock = true;
+  }
+
+  if (version.prerelease.length && prereleaseLock) {
     // Find the set of versions that are allowed to have prereleases
     // For example, ^1.2.3-pr.1 desugars to >=1.2.3-pr.1 <2.0.0
     // That should allow `1.2.3-pr.2` to pass.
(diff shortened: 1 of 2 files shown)


protyposis:
Package is available from the NPM repo as [`semver-upr`](https://www.npmjs.com/package/semver-upr).

climblinne:
Is there any chance to get this into `master`? 
It would be nice to get it also e.g. into `max_satisfying` function.

isaacs:
This is a good idea, and addresses a real need, but I definitely feel like the maximum number of boolean options is 1 (or maybe zero), so we need to switch to an options object.  Also, the default behavior should be used when the option is falsey, for consistency.

In order to prevent a backwards compatibility breakage and major version bump, we can add `if (options === true) options = { loose: true }` in the functions currently taking a loose param.  Then add `includePrerelease` as a flag, and when that's truthy, all prerelease versions are included in any range based strictly on precedence ordering.

isaacs:
(Don't feel pressure to do extra work.  If you don't get to it, I will relatively soon.  Just documenting the decisions here so it's not left hanging, if anyone else wants to pick it up.)

isaacs:
Actually I lied, I had a meeting get canceled so I did it.

The question now is whether this is semver major or minor.  It _does_ change the case when you might've used a truthy non-boolean as the `loose` param.

climblinne:
Thanks a lot. I think, I would go for a major change.

protyposis:
The docs said

> All methods and classes take a final `loose` **boolean** argument

so if someone provided a truthy value instead of `true`, he'd have used the API wrongly and just gotten lucky that it worked. Viewed like this, it would be a minor change. Since this library is used by quite a lot of people, it might make sense to take the safe route though and make it major.

Anyway, thanks for integrating this 👍

isaacs:
The only truthy value that would cause an issue (in the current code on master) would be an object without a `loose` field.  If you pass in any non-object truthy value, it'll explicitly coerce to `{ loose: true }`.

PR #195, opened 2017-05-16
calculate two Ranges intersection.

xice-blue:
I have two case need deal:
3 modules dependency X module:
1.
'x': '~1.1.0'
'x': '~1.1.5'
'x': '1.1.10'
or
2.
'x': '~1.1.0'
'x': '~1.1.5'
'x': '1.2.1'
Case 1 need get X module version is 1.1.10, and case 2 throw Exception about version mismatching.
So I extend function 'intersections' to deal these cases.


rtfpessoa:
I think this is the same I have here: #187 waiting for merge for a while.

xice-blue:
@rtfpessoa your function return is true or false, I need return intersection.
For example: range '~1.1.0' and '^1.0.0 < 1.1.5', it will return '>=1.1.0 <1.1.5'.
And I can get more range's intersection.

rtfpessoa:
You are right. I think I misread the description.

v3lbloud:
Please add test for `['^1.2.1', '^1.2.5', '>=1.2.5 <2.0.0']`. It doesn't seem to be working.

isaacs:
Landed and published in v5.4

isaacs:
Er, another implementation landed.  Not this one.

PR #200, opened 2017-07-08
range.bnf: Remove meaningless pipe

motet-a:


isaacs:
That's not meaningless.  It means "or nothing" in this context.

motet-a:
Oops! Sorry, I really thought it was an error.

(By the way, I am probably bikeshedding but what about replacing `( '<' | '>' | '>=' | '<=' | '=' | )` by a more explicit `( '<' | '>' | '>=' | '<=' | '=' | '')` with an empty string at the end?)

motet-a:
Moreover, `primitive ::= ( '<' | '>' | '>=' | '<=' | '=' | '') partial` seems to me equivalent to `primitive ::= ( '<' | '>' | '>=' | '<=' | '=') partial` since `primitive` is only used in the `simple` rule, which is `simple ::= primitive | partial | tilde | caret`.

I think it should be either

```js
simple     ::= primitive | partial | tilde | caret
primitive  ::= ( '<' | '>' | '>=' | '<=' | '=') partial
```

or

```js
simple     ::= primitive | tilde | caret
primitive  ::= ( '<' | '>' | '>=' | '<=' | '=' | '') partial
```

isaacs:
Oh!  Right, because it gets composed with a thing that also includes `partial`, that's a redundant entry.

isaacs:
Landed with a slight commit message edit in 7184ff429eaf6a969b292d4129e4fffeb666031d.

PR #208, opened 2017-09-11
add minVersion function

kellyselden:
I'm trying to decide which range is "newer" when automatically updating package.json [here](https://github.com/kellyselden/ember-cli-update/pull/108) and [here](https://github.com/bantic/three-way-merger). The scenario is

* my dep is "a": "^2.0.2"
* upstream was updated "a": "^1.4.0" => "a": "^2.0.0"

I want to be smart and say, even though the dep was updated, mine is more recent, so ignore the change. My pseudocode is `if myDepRange < newDepRange then update`.

Getting the lowest possible version in a range seems like a good way to compare two ranges for newest. I currently am doing:

```js
function minVersion(range) {
  return new Range(range).set[0][0].semver.version;
}
```

But I feel like I'm reaching into too much private API. I would like add this function upstream to this library.

### DIFF
--- a/semver.js
+++ b/semver.js
@@ -1200,6 +1200,19 @@ function validRange(range, loose) {
   }
 }
 
+exports.minVersion = minVersion;
+function minVersion(range, loose) {
+  try {
+    var rangeObj = new Range(range, loose);
+  } catch (er) {
+    return null;
+  }
+  if (!rangeObj.range) {
+    return '0.0.0';
+  }
+  return rangeObj.set[0][0].semver.version;
+}
+
 // Determine if version is less than all the versions possible in the range
 exports.ltr = ltr;
 function ltr(version, range, loose) {
--- a/README.md
+++ b/README.md
@@ -338,6 +338,7 @@ strings that they parse.
 ### Ranges
 
 * `validRange(range)`: Return the valid range or null if it's not valid
+* `minVersion(range)`: Return the lowest version that satisfies the range.
 * `satisfies(version, range)`: Return true if the version satisfies the
   range.
 * `maxSatisfying(versions, range)`: Return the highest version in the list
--- a/test/index.js
+++ b/test/index.js
@@ -695,6 +695,22 @@ test('\nstrict vs loose ranges', function(t) {
   t.end();
 });
 
+test('\nmin version', function(t) {
+  [['1.2', '1.2.0'],
+    ['>=2.1.0 <2.2.0', '2.1.0'],
+    ['', '0.0.0'],
+    ['~3.3', '3.3.0', true],
+    ['invalid range', null]
+  ].forEach(function(v) {
+    var range = v[0];
+    var expect = v[1];
+    var loose = v[2];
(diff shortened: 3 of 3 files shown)


jaydenseric:
I would really like this too, to be able to derive the minimum supported Node.js version from the range in `package.json` `engines`. This is useful for configuring `@babel/preset-env` to transpile and polyfill for the minimum supported Node.js version.

To easily [do this in my own packages](https://github.com/jaydenseric/apollo-upload-client/blob/v7.0.0-alpha.4/.babelrc.js#L10) I make the assumption the range is always a single semver prefixed with `>=`, so I can just strip it off with `substring(2)` to get a min version.

[`@babel/preset-env` will eventually look at `engines.node` to do this automatically](https://github.com/babel/babel/issues/7277), so a more robust solution that can take arbitrary, complicated ranges will be useful.

semver.js
```diff
@@ -1200,6 +1200,19 @@ function validRange(range, loose) {
...
+function minVersion(range, loose) {
+  try {
+    var rangeObj = new Range(range, loose);
+  } catch (er) {
+    return null;
+  }
+  if (!rangeObj.range) {
+    return '0.0.0';
+  }
+  return rangeObj.set[0][0].semver.version;
```
isaacs: I think this will return an incorrect value in the following cases:

```
minVersion('1.2.3 || 1.0.0') // expect 1.0.0, get 1.2.3
minVersion('<=1.2.3 >1.0.0') // expect 1.0.1, get 1.2.3
minVersion('>1.2.3') // expect 1.2.4, get 1.2.3
minVersion('>1.2.3-alpha') // <-- this one is probably impossible? 1.2.3-alpha is incorrect tho
```

isaacs:
I think that this might be theoretically impossible.  You could get the minimum _of a given set of versions_ that satisfies a given range.  But what you can't do is (usually) specify the minimum version that would satisfy a range.  Consider `>1.2.3-alpha` which would be satisfied by `1.2.3-beta` or `1.2.3-alpha.0` or `1.2.3-alphb`.

jaydenseric:
The [`no-unsupported-features` rule of `eslint-plugin-node`](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features.md#rule-details) would probably benefit from this too. It seems they [changed their approach](https://github.com/mysticatea/eslint-plugin-node/pull/99) from attempting to determine a min version.

semver.js
coreyfarrell: I was working on a patch for this before I saw this PR. #241 deals with these edge cases and includes testing for them situations.  `minVersion('>1.2.3-alpha')` is not impossible, I believe the correct result is `1.2.3-alpha.0` which is less than `1.2.3-beta`.

coreyfarrell:
`semver.minVersion` was implemented by #241 and released in semver@5.7.0.  Sorry for bypassing this PR, I created/submitted the patch before I saw this.

PR #215, opened 2017-11-07
Added version coercion to the module and CLI.

joeledwards:
Added the ability to coerce a string to semver before validation. This applies both to the CLI (via the -c,--coerce option) and the module (via the coerce function).

### DIFF
--- a/semver.js
+++ b/semver.js
@@ -21,6 +21,9 @@ exports.SEMVER_SPEC_VERSION = '2.0.0';
 var MAX_LENGTH = 256;
 var MAX_SAFE_INTEGER = Number.MAX_SAFE_INTEGER || 9007199254740991;
 
+// Max safe segment length for coercion.
+var MAX_SAFE_COMPONENT_LENGTH = 16;
+
 // The actual regexps go on exports.re
 var re = exports.re = [];
 var src = exports.src = [];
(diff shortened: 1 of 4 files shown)


semver.js
```diff
@@ -1294,3 +1294,24 @@ function intersects(r1, r2, loose) {
...
+  if (version instanceof SemVer)
+    return version
+
+  if (typeof version !== 'string')
+    return null
+
+  if (version.length > MAX_LENGTH)
+    return null;
+
+  var match = version.match(/([vV]?\d+(?:[.]\d+){0,2})/)
```
isaacs: Most of the rest of this looks fine.  Just reviewing the regexp.

- Seems like the `[vV]` is unnecessary, since it's optional, and being thrown away anyway.
- The `\d+` should be `\d{1,256}` to prevent ReDOS attacks by passing in a string of a million digits.  (Docs should probably say that only the first `n` digits will be captured, so `1{1e10}.2.3` will be coerced to `1{256}.0.0`.)  I don't know if 256 is the right number, but it should match what's supported elsewhere.
- I'd use `\.` rather than `[.]`.  Arbitrary stylistic nit, but that's how it's done elsewhere.
- This regexp should be defined up with the others at the top of the file.  (Again, arbitrary stylistic nit, just a handy way to keep things manageable in a module with a gillion fiddly regexps.)
- If the three blobs were capture groups, then the re-assembly would be easier later (no join or array walk, just 3 conditional `||` assignments.)

I think that works out to:

```
/(\d{1,256})(?:\.(\d{1,256}))?(?:\.(\d{1,256}))?/
```

(but check on the 256 value, I'm not sure that's correct.)

Then the coerced version is:

```
parse((match[1] || '0') + '.' + (match[2] || '0') + '.' + (match[3] || '0'))
```

isaacs:
Cool, lgtm!

PR #217, opened 2017-11-13
Clarify Caret Ranges

lm1:


joeledwards:
I think the term `value` is clearer than `item`. All for replacing `digit`.

lm1:
I thought 'tuple value' could be a bit ambiguous, but it should be clear it's not referring to entire tuple in this context.

joeledwards:
Perhaps 'element'?

isaacs:
e09ecf8 thanks!

Issue #232, opened 2018-03-02
Coerce returns null if have a 0 appended to major, minor or patch


lomby92:
Using **semver.coerce(<string>)** returns **null** value if there is one or more '0' before _major_, _minor_ or _patch_ value.
Some example of versions that fail:
- '0.02'
- '1.04'
- '0.67.09'
- '2018.06.05'

Maybe the coerce code can check also _major_, _minor_ and _path_ values as final step.

lomby92:
@isaacs i have a fix for that, how can i push? Now i have permission denied

bartvanandel:
The problem is not restricted to major, minor or patch value, it also fails with `2.0.0-20171117T112133.0734735.0`, i.e. a dot-separated part of the prerelease tag (which in our case is a git hash). It seems this problem only occurs when the part contains only numeric characters and starts with a zero (or zeroes). Adding a letter somewhere or removing the initial zeroes results in the expected result:

```
λ node
> const semver = require('semver')
undefined
> semver.valid('2.0.0-20171117T112133.0734735.0')
null
> semver.valid('2.0.0-20171117T112133.x734735.0')
'2.0.0-20171117T112133.x734735.0'
> semver.valid('2.0.0-20171117T112133.734735.0')
'2.0.0-20171117T112133.734735.0'
```

bartvanandel:
@lomby92: if you create a fork of this repo and put your fix in a branch there, you can create a Pull Request from that branch :)

andytompkins:
I just ran into this today. I was quite surprised that '1.123.000' was not considered valid, and that it would not coerce either. Sure, its redundant and silly that I have 000, but its actually outside of my control (I don't version the component that returns that.) It would be good if this could get fixed!

yedlosh:
It's also then considerably misleading to say in readme that: 
"Only text which lacks digits will fail coercion"

lukekarrys:
This can be done by using `loose: true` currently:

```
> semver.coerce('0000.004.02', { loose: true }).version
'0.4.2'
```

PR #233, opened 2018-03-05
Add leading zeros remove

lomby92:
Coerce method can remove leading zeros from minor, major and patch.
EG: '0.004.02' -> '0.4.2'

### DIFF
--- a/semver.js
+++ b/semver.js
@@ -1307,6 +1307,31 @@ function intersects(r1, r2, loose) {
   return r1.intersects(r2)
 }
 
+function removeLeadingZeros(num){
+  let countLeadingZeros = 0
+  let chars = num.split('')
+  let foundNonZeroChar = false
+  // iterate chars until first non-zero char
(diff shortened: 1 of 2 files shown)


lomby92:
Ok @bartvanandel , i will edit as you suggested

erichs:
@isaacs this is still an active issue. Can this conflict be resolved and the patch applied?

lomby92:
@isaacs 
I can update the PR, let me know if this feature could be useful

yedlosh:
@bartvanandel Could this be merged in to finally fix #232?

bartvanandel:
I don't think I can merge this, I don't own this repo.

Vlad-AndreiMunteanu:
Is this going to get merged at all? I face the same issue with leading zeros today.

lukekarrys:
This can be done by using `loose: true` currently, but it can't be made the default in order to be spec compliant.

```
> semver.coerce('0000.004.02', { loose: true }).version
'0.4.2'
```

Issue #228, opened 2017-12-25
Feature request: A way to determine if range1 is subset of range2.


coreyfarrell:
I'm working with lint tests which look at package.json and throw error's if unavailable features are used.  For example if package.json contains `"engines": { "node": ">=0.10" }` but the code uses arrow functions, an error is throw.  The basic implementation is:
```js
if (semver.intersects('>=0.10', '<4')) {
  throw new Error('package.json should not allow node.js <4');
}
```
This works but now I'd like to test dependencies.  For example if my package says it's compatible with `node>=0.10` but requires a package that says `node>=4`, I want to create a rule that will report this as an error.  So:
```js
if (!semver.subset('>0.10', '>=4')) {
  throw new Error('The local package advertises compatibility with 0.10 but deps require 4.0.0')
}
```
In this example `>0.10` is not a subset of `>=4` because it includes `>=0.10 <4` so semver.subset returns false.  I'm indifferent to the parameter ordering, maybe the complete set should be first with the subset to be tested second?  In any case the goal is the ability to verify that one semver range falls completely within another.

coreyfarrell:
semver.minVersion from #241 actually resolved my reason for posting this issue.  My goal is to be able to look at `engines.node` from package.json and determine if it excludes all versions of node which do not support a specific feature, mainly so eslint rules can auto-disable if `engines.node` is set to old for the rule to apply.

@righthandpaw do you intend to continue working #235?  If not I plan on just closing this issue.

PR #235, opened 2018-03-30
Added Subset Feature - A way to determine if range1 is subset of rang…

righthandpaw:
Subset Feature - Determine if range1 is subset of range2. #228

https://github.com/npm/node-semver/issues/228

### DIFF
--- a/semver.js
+++ b/semver.js
@@ -1307,6 +1307,50 @@ function intersects(r1, r2, loose) {
   return r1.intersects(r2)
 }
 
+exports.subset = subset;
+function subset(r1, r2, loose){
+	
+  if(r2.match(/^</)) r2 = "0 "+r2;
+	
+  r1 = new Range(r1, loose); 
+  r2 = new Range(r2, loose); 
+	
+  set_a = r1.range.match(/([0-9\.])+/g).sort(function(a, b){
+    a = a.match(/[0-9]+/);
+    b = b.match(/[0-9]+/);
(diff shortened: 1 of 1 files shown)


isaacs:
This has some tricky edge cases that aren't handled here, because a "range" can be much more than just a single comparator.

Consider the range `5.0.0 || >=1.2.3 <5.0.0 || 6.7.8`.  Is `1.2.4 - 5.1.0` a subset of that?  (The answer is no.)  Is `1.2.4 - 5.1.0 || 6.7.8`?  (Yes.)

This function would have to handle cases like this.  Rather than string manipulation, I'd approach it like this:

1. Turn both arguments into Range objects.  These objects have a set of arrays of Comparator objects.  (A 2-d array of them.)
2. Verify that, for each set of comparators in r1, there exists at least one comparator array in r2 that contains all comparators in the r1 set array.

If each of the comparator sets in r1 are a subset of at least one of the comparator sets in r2, then r1 is a subset of r2.

coreyfarrell:
I'm not getting expected results:
```js
subset('>=4', '>=6'); // false
subset('>=6', '>=4'); // false
```

I expected one of them to be true since >=6 is a subset of >=4.  To be honest I've confused myself about the order of arguments but those two calls should have opposite returns.

isaacs:
A different `semver.subset()` method was landed in 7.3.  Closing this.

PR #244, opened 2018-07-04
Update tap version

bzoz:
This makes tests pass on Windows. Currently it is broken because of https://github.com/tapjs/node-tap/issues/373

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,25 +4,31 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "ajv": {
+      "version": "5.5.2",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-5.5.2.tgz",
+      "integrity": "sha1-c7Xuyj+rZT49P5Qis0GtQiBdyWU=",
+      "dev": true,
+      "requires": {
+        "co": "^4.6.0",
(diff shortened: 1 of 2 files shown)


jdalton:
@isaacs This PR would also allow unskipping Windows tests for `semver` in Node's CitGM.

jdalton:
Thank you @isaacs!

Would it be helpful if I pop-in to similar issues for other packages you maintain and identify them as related to CitGM? _(this is related to an MS effort to improve Node on Windows)_

isaacs:
Thanks!  I actually bumped it all the way to 12.0.1, but effect should be the same.  Do you need a push to npm as well, or does citgm pull from the git repo?

isaacs:
@jdalton Yes, please, if there's anything I or npm can do to help make node on windows better, definitely feel free to ping me.  Making citgm better is always winful.

jdalton:
@isaacs

> Do you need a push to npm as well, or does citgm pull from the git repo?

Yes. CitGM references the latest version number published to npm and maps it back to a GitHub tag in the package's repo to get its tarball for testing.

> Yes, please, if there's anything I or npm can do to help make node on windows better, definitely feel free to ping me. Making citgm better is always winful.

Thank you, will do!

PR #247, opened 2018-07-19
Fix grammar for ranges

steveklabnik:
This *might* be wrong, so please let me know!

I took the `| ` in this line to be the "with no operation" case, but the rule right above has

    simple     ::= primitive | partial | tilde | caret

so, the "no operator" case is already covered. This makes me think the `|  ` was a typo, so it should be removed.


isaacs:
It's not a typo per se, but it is extraneous, and this is clearer. 👍

steveklabnik:
❤️👍

PR #252, opened 2018-07-30
Clarify some ^0.0.x & ~.0.0.x cases (TBD in test only)

brodybits:
Adding some positive & negative test cases, as triggered by <https://npm.community/t/wanted-version-is-lower-than-what-it-should-latest>. (It did not seem so clear to me from <https://docs.npmjs.com/misc/semver#tilde-ranges-123-12-1> & <https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004>).

TODO I think this really needs to be documented. I think it is not super tricky if tested and documented properly. I hope to have a proposal really soon.


isaacs:
93c2f2d thanks!

Issue #224, opened 2017-12-09
Incrementing "1.0" does not work, returns undefined


hansemannn:
Hey guys!

We noticed that when specifying a version like `1.0` instead of `1.0.0` to `semver -i` / `semver.inc()`, it does not increment it properly and returns undefined. Example:

```
$ semver 1.0.0 -i major
2.0.0
$ semver 1.0 -i major

```

Is there an option to bump it anyway or do we need to manually repair it? It feels overhead to do it but would of course be possible as well. Thank you!

isaacs:
1.0 isn't a "version", it's a "range".  (That is, it's equivalent to `1.0.x`, and matches any non-prerelease versions from `1.0.0` up to but not including `1.1.0`.)

A better error message and non-zero exit code would be great, though.

Issue #146, opened 2015-12-07
Standard?


ekmartin:
Right now `bin/semver` has a style that's quite different from the rest of the code, and since npm core uses standard maybe it would be a good idea to use it for this project too? If you think so I'd be glad to send a pull request with the necessary changes.

PR #255, opened 2018-10-10
Fix bin/semver runtime error

pjohnmeyer:
Changes since v5.5.1 have introduced an error when running the
semver CLI:

`ReferenceError: options is not defined`

This commit fixes this problem, then adds a few sanity tests
that run the CLI to hopefully ensure this doesn't happen again.

### DIFF
--- a/bin/semver
+++ b/bin/semver
@@ -17,6 +17,7 @@ var argv = process.argv.slice(2)
   , identifier = undefined
   , semver = require("../semver")
   , reverse = false
+  , options = {}
 
 main()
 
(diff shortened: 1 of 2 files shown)


pjohnmeyer:
I make this pull request because I was looking at working on #224 and possibly #146, but ran into this issue.

I think it might make sense to, in a future PR:

* Move the actual logic of the CLI into a different requirable file, like `semver-cli.js`, so it can be tested without using `execSync`.
* Change `bin/semver` to a wrapper that basically just does something like the below, but with additional error handling:
```js
var argv = process.argv.slice(2);
var main = require('../semver-cli').entryPoint;
main(argv);
```

PR #257, opened 2018-10-11
test:💥 something amazing is in the works 💥

bcoe:


.travis.yml
```diff
@@ -1,11 +1,18 @@
 language: node_js
-sudo: false
```
XhmikosR: IIRC you could keep this here @bcoe

PR #258, opened 2018-10-12
Add 'standard' to npm test and address issues

### LINKED ISSUE #146: Standard?

pjohnmeyer:
Resolves #146

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,10 @@
   "description": "The semantic version parser used by npm.",
   "main": "semver.js",
   "scripts": {
-    "test": "tap test/*.js --cov -J"
+    "test": "standard && tap test/*.js --cov -J"
   },
   "devDependencies": {
+    "standard": "^12.0.1",
     "tap": "^12.0.1"
   },
   "license": "ISC",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "The semantic version parser used by npm.",
   "main": "semver.js",
   "scripts": {
-    "test": "standard && tap test/*.js --cov -J"
+    "test": "standard bin/semver *.js **/*.js && tap test/*.js --cov -J"
```
isaacs: Can you move the standard command invocation to a `posttest` script instead?

I prefer to run tests _before_ linting, so that it's less annoying when building new features or fixing bugs, and the `&&` is a sh-ism that doesn't work on Windows.

pjohnmeyer: Makes sense to me, working on it now.

isaacs:
Rebased and landed on master.  Thanks!

PR #262, opened 2018-11-28
Feature subset

GitNiko:
Determine if range1 is subset of range2:
```js
semver.subset(semver.Range(subset1), semver.Range(subset2))
```


GitNiko:
Any updates on merging this? @isaacs

underscorediscovery:
Some notes I had from recently playing around with this:

Add tests for any ranges: for example, `["*", "*", true]` might cause issues in this implementation, as it touches `semver.version` which is assigned `ANY = {}`.
Since these cases are well defined they can be shorted up front to skip the work.

- any/any - `*`, `*` - always `true`
- any/other - `*, inner` - always `true`
- other/any - `outer, *`  - always `false`

I also would suggest clearer naming, `subset(outer, inner)` is a lot clearer to understand which argument to put into which position. If on the range class (as an instance method) it can be clearer as `subset` => `allowsAll` and `intersects` => `allowsAny`.

GitNiko:
@underscorediscovery thanks for your suggestion😄, I will update it.

underscorediscovery:
@GitNiko sure. btw I also found a case that fails for this implementation,`["1.0.0", "<1.1.0", true]`.

GitNiko:
@underscorediscovery has been fixed.

isaacs:
This feature was implemented in a slightly different way on master.  Thanks!

PR #271, opened 2019-03-09
Range intersect supports wildcards

Asthetic:
- Added undefined check for range operator, wildcards defaults to empty.
- Comparator and Range test returns true if either version is a wildcard


Asthetic:
I just thought of something, what if we changed wildcard to translate to >=0.0.0? What impact would that have on other code? It might make it easier to use intersects for comparators.

Issue #239, opened 2018-06-06
`intersects` returns true for some ranges that do not intersect


corhere:
The `intersects` function can incorrectly return `true` when the first range has a comparator set with multiple comparators, and the second range has more than one comparator set (`a || b`).

```js
intersects('1.5.x', '<1.5.0')  // false
intersects('1.5.x', '>=1.6.0')  // false
intersects('1.5.x', '<1.5.0 || >=1.6.0')  // true?
intersects('<1.5.0 || >=1.6.0', '1.5.x')  // false
```

The flaw in the `intersects` algorithm is that it considers a set in the first range to intersect if every comparator in the set intersects with every comparator in **any** set in the second range. It needs to instead test whether there is some set in both ranges where all the comparators intersect.

isaacs:
Fixed in version 6.0.0.  (It's a breaking change, even though it's a bugfix, because it could cause programs to need to be updated to keep working properly.)

PR #274, opened 2019-03-21
fix: Fix Range intersects algorithm

max-b:
Fixes https://github.com/npm/node-semver/issues/239

I think there's still an open question about whether `<1.0.0 >2.0.0` and `>0.0.0` intersect because technically `<1.0.0 >2.0.0` is an empty set and so can't intersect with anything, but that would require a more complicated solution and this PR is a more trivial fix.


isaacs:
Yeah, `<1.0.0 >2.0.0` should not intersect `>0.0.0`.

Intersect should not be "is the first set unchanged by adding the second set", it's "is there a version somewhere that satisfies both ranges".

max-b:
Ok I think that could be addressed with something like:
```
  return this.set.some(function (thisComparators) {
    return range.set.some(function (rangeComparators) {
      return isSatisfiable(thisComparators) && thisComparators.every(function (thisComparator) {
        return isSatisfiable(rangeComparators) && rangeComparators.every(function (rangeComparator) {
          return thisComparator.intersects(rangeComparator, options)
        })
      })
    })
  })
```
Where `isSatisfiable` would do essentially what it says on the lid.

Would you consider this a breaking change? It's technically a fix, but it will require changing test results and it's possible that there's someone out there depending on the previous behavior/idiosyncrasy?

max-b:
Great! I've been using semantic-release style commits out of habit, but it doesn't really look like they're being fully used here (and I think none of them would actually trigger a major version change...). I'd imagine that if everything looks ok you can decide how'd you like to merge/rebase it.

Thanks!

semver.js
```diff
@@ -799,11 +799,15 @@ Range.prototype.intersects = function (range, options) {
 
   return this.set.some(function (thisComparators) {
     return range.set.some(function (rangeComparators) {
-      return isSatisfiable(thisComparators, options) && thisComparators.every(function (thisComparator) {
-        return isSatisfiable(rangeComparators, options) && rangeComparators.every(function (rangeComparator) {
-          return thisComparator.intersects(rangeComparator, options)
+      return (
+        isSatisfiable(thisComparators, options) &&
```
isaacs: This could also be moved up a line.  No sense walking `range.set` if `thisComparators` isn't satisfiable.

max-b:
Yay 🎉 Thanks for the feedback/hand-holding 😬 

Should be able to close #239 now 😄

isaacs:
This landed on v6.

PR #284, opened 2019-07-12
export new RELEASE_TYPES constant

hcharley:
This will help library consumers better integrate semver into their apps.

Possible usecases of consuming this constant:

- Automatically show a dropdown of available release types that conform to the semver spec
- Validate user inputs where a release type is expected

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,3 +441,39 @@ ex.
 * `s.clean('  =v2.1.5')`: `2.1.5`
 * `s.clean('      2.1.5   ')`: `'2.1.5'`
 * `s.clean('~1.0.0')`: `null`
+
+
+## Constants
+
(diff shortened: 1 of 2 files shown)


charlex:
I am happy to add this to the docs if this idea is approved by the maintainers.

isaacs:
Yes, please add to the docs, and add at least a basic test.  (Just verify it exists and is what you expect; I realize it's kinda silly for such a trivial feature, but it prevents us from accidentally breaking it in the future and not noticing.)

charlex:
@isaacs  I added a section in the README, but it looks like this project as undergone a big change in its structure, and I'm not able to resolve the conflicts right now.

wraithgar:
Replaced by #530

PR #295, opened 2019-11-18
Isolate functions so they can be tree-shaken

### LINKED ISSUE #291: [FEATURE] Separate functions into isolated imports for tree shaking
# What / Why
We rely on two or three functions out of this package, and it costs us 5kbs minified at runtime in the browser. We would like the ability to treeshake modules such that we can continue to rely on the package.

Here is the bundlephobia for the package currently. https://bundlephobia.com/result?p=semver@6.3.0

## Where
Enterprise products owned by Intuit

## How
Instead of doing

```js
import semver from "semver"

// ...

semver.valid(/* ... */);
```

Have the ability to do

```js
import valid from "semver/valid"

valid();
```

### Current Behavior
I must import the entire default module to ensure that I can get a single function off of the module.

### Expected Behavior
I should be able to import only the functionality that I care about when targeting the web.

## Who
@aghassi

## References
* n/a

Although there are similar alternatives for specific use cases such as https://www.npmjs.com/package/semver-compare

stabbylambda:
# What / Why
Resolves #291. (For reference, @Aghassi and I work together) Basically we want to tree shake out stuff we don't use when comparing version ranges in the browser.

# So what did I do?

Basically took everything in semver.js and moved it into separate files. The structure now looks like
```
/classes        // SemVer, Range, and Comparators
  semver.js
  ranges.js
/internal       // stuff nobody should import, ever
/functions      // the individual functions for an individual SemVer
  clean.js
  cmp.js
  coerce.js
  ...
/ranges         // the individual functions for Ranges
  gtr.js
  intersects.js
  ltr.js
  ...
/index.js       // the original semver.js file moved to index.js and re-exports the exact same interface
```

# What does the API look like now?

Original use continues as-is, since the public api of semver.js (now index.js) didn't change.

However, we now have some nice tree-shaking when we do:

```js
import valid from "semver/functions/valid"
valid("2.0.0");
```

## What else

I briefly attempted to split Comparator and Range from each other in `ranges/index.js` but since they reference each other, it seemed a little pointless, as they'll both be pulled into the browser when required anyway. (Not to mention our code doesn't use them.)


index.js
```diff
@@ -0,0 +1,53 @@
+var SemVer = require('./SemVer');
+var { compareIdentifiers, rcompareIdentifiers } = require('./internal/identifiers');
+var { SEMVER_SPEC_VERSION } = require('./internal/constants');
+var { re, src, t } = require('./internal/re');
```
isaacs: Doing object dereferencing assignment requires a more recent node version than let/const and classes.  May as well update to let/const, classes, and arrow functions in this change as well, and we can make it require node 10 and higher.

internal/debug.js
```diff
@@ -0,0 +1,17 @@
...
+  debug = function () {
+    var args = Array.prototype.slice.call(arguments, 0)
+    args.unshift('SEMVER')
+    console.log.apply(console, args)
+  }
+} else {
+  debug = function () {}
+}
+
+module.exports = debug;
```
isaacs: No unnecessary semicolons, please.

index.js
isaacs: Alternatively, if you think it's worth keeping the existing node version support for this change, remove all the object dereferencing.

isaacs:
Also, would it maybe make sense to have a `./class` folder containing the SemVer, Range, and Comparator definitions?

index.js
stabbylambda: I changed:

- var -> const/let (where appropriate)
- SemVer, Range, Comparator -> classes (but I left the helper functions for Range outside the class since they're sorta kinda private
- functions -> arrows _in the callbacks_. I didn't change the named functions in `/functions` or `/ranges` mostly because then I'd have to change it like below in order to keep the names:
```js
module.exports = function compare(...) { ... }
//to
const compare = (...) => {...}
module.exports = compare
```
- string concat -> template strings. I know you didn't request it, but it stuck out like a sore thumb once I had all the other es2015+ stuff in there 😄

stabbylambda: Oh I also fixed all the casing stuff. 👍

stabbylambda:
I'll also take a swing at the tests. I mostly wanted to leave them alone while I was refactoring the code because...you know. 💥

stabbylambda:
@isaacs Alright. I think this is it. I left all the eq/neq/lt/gt/cmp tests in index.js because they use the same data and I really didn't want to break all that up. But I think everything else is where it needs to be. 👍

test/functions/compare-loose.js
```diff
@@ -1,6 +1,6 @@
 const { test } = require('tap')
 const compareLoose = require('../../functions/compare-loose')
-const SemVer = require('../../classes/SemVer')
+const SemVer = require('../../classes/semver')
```
isaacs: Called it!  :P

stabbylambda: 😄 Stupid macOS and its permissive casing.

isaacs:
So, I squashed this and carried the ball a bit further, in #296.

This PR will close automatically when that one lands.  If you feel like adding tests for the classes, you can send a PR against that branch.

The Range and Comparator classes each have their own standalone function, and the `max-satisfying` method is moved into `functions/`.  Also, the index lazily loads the components that are used, which should make things a bit more performant in some cases, so npm doesn't have to load methods that it never calls.

PR #301, opened 2019-12-17
Add preload.js into NPM package

Flarna:
# What / Why

preload.js is missing in NPM package as it is not mentioned in files section of package.json

## References

* n/a


isaacs:
Oh, whoops.  THanks!

PR #302, opened 2019-12-22
chore: minor code refactoring

jamesgeorge007:
Replaced `var` with `let`

### DIFF
--- a/functions/diff.js
+++ b/functions/diff.js
@@ -8,9 +8,10 @@ const diff = (version1, version2) => {
     const v1 = parse(version1)
     const v2 = parse(version2)
     let prefix = ''
+    let defaultResult
     if (v1.prerelease.length || v2.prerelease.length) {
       prefix = 'pre'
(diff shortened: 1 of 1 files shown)


functions/diff.js
```diff
@@ -8,9 +8,10 @@ const diff = (version1, version2) => {
...
     const v2 = parse(version2)
     let prefix = ''
+    let defaultResult
     if (v1.prerelease.length || v2.prerelease.length) {
       prefix = 'pre'
-      var defaultResult = 'prerelease'
+      defaultResult = 'prerelease'
     }
     for (const key in v1) {
       if (key === 'major' || key === 'minor' || key === 'patch') {
```
jamesgeorge007:
```js
if (['major', 'minor', 'patch'].includes(key)) {
  //
}
```
Maybe :thinking:

isaacs: Three `===` checks is fine.

PR #308, opened 2020-01-11
DRY refactor

fisker:
# What / Why

> For easier maintaince.

## References

* n/a


isaacs:
Thanks, but probably not going to move in this direction.  Better for `require()` to only ever load string values, for webpack usage.

PR #311, opened 2020-01-22
Remove the fancy preload logic in index.js

### LINKED ISSUE #306: [BUG] Upgrading to version 7 causes webpack error
We're currently using semver 6.3.0, but if we upgrade to the latest version (7.1.1), we start getting webpack errors:

```
WARNING in ./node_modules/semver/index.js 3:51-64
Critical dependency: the request of a dependency is an expression
```

Based on the usage notes in the README, it would seem that we now need to use `require` with `semver/preload`? Why?

This is how we've been importing from semver for a while now, but this doesn't work anymore:

```
import { coerce, gte } from 'semver';
```

If I try to import from `semver/preload`, then I start getting TypeScript errors because that string doesn't match a known module name. What's the correct `import` API syntax here?

isaacs:
This is causing some friction for webpack users, and doesn't seem to  be
worth the trouble from a performance point of view in npm/cli.

Ditch it.

Fix #306


isaacs:
> Does any documentation need to change with the landing of this pull-request?

Good call, there was one weird nit.  Fixed and force-pushed.  I'll pull this and publish now.

PR #320, opened 2020-04-07
Exclude prereleases from range maximums

### LINKED ISSUE #223: `intersects`: claims that `15` and `^16.0.0-0` intersect, but they should not
(this necessitated [this workaround](https://github.com/airbnb/enzyme/pull/1404))

Basically, `semver.intersects('15', '^16.0.0-0')` returns `true`, and should return `false`. `semver.intersects('15', '^16.0.0')` properly returns `false`, and https://semver.npmjs.com/ for "react" and "^16.0.0-0"  properly does not include `15`.

I'm happy to submit a PR for this soon.

### LINKED ISSUE #254: `semver.intersects` behaves inconsistently comparing range with prerelease version vs range
I know prerelease versions have a bit of edgecasey behavior, but I'm hoping to nail down if `false` is the unexpected result, or `true` is, or if this is just expected all around, given
```
semver.intersects(`<7.0.1`, "7.0.0-beta.0") // false - surprising?
semver.intersects(`<7.0.1`, "^7.0.0-beta.0") // true - expected
```

The inconsistency here is because [`.intersects` falls back to `.satisfies`](https://github.com/npm/node-semver/blob/acb44ddf2aef5855a1eb77fedc6b2e2c0868f7c8/semver.js#L728) when one of the arguments is an exact version, and `.satisfies` has [all this extra stuff](https://github.com/npm/node-semver/blob/acb44ddf2aef5855a1eb77fedc6b2e2c0868f7c8/semver.js#L1128) that doesn't apply when comparing a range with a range.

I can't tell if this is actually a bug in `.satisfies`, or what.

### Background

Babel is trying to use `semver` programmatically to decide when certain versions of things are available, and we ran into a little edge case. I _think_ we've worked around it for now with https://github.com/babel/babel/commit/13798feefbf19389cb692aa0b5157276ce7cdf4f#diff-7a78d3752af461cdbf66d58d59a2026dR169, but one thing I'd like verified, if possible, is whether this fix is actually safe. If the second (`true`-returning) example is actually incorrect, then a fix for it would cause serious regressions in behavior for Babel users, for instance.

eemeli:
In semver v4, the ranges were simplified in 2331a9e "to remove -0 everywhere". This had the unintended side effect of including the endpoint version's prerelease versions in the resulting range, as `semver.lt('2.0.0-0', '2.0.0') === true`.

This bug is apparent in intersection tests:
```js
semver.intersects('1', '2.0.0-0') // false
semver.intersects('1', '^2.0.0-0') // was true, here fixed to false
```

To fix this, the `Range` class is updated to always include `-0` in the `<` Comparator when generating one from expressions like `1.x`, `1.0 - 1.5`, `~1.2.3`, or `^1.2`.

Tests are updated, moving one test from range-include to range-exclude, comparing `2.0.0-rc1` against `^1.0.0` when `includePrerelease` is true

Fixes #223
Fixes #254


elashpear:
@isaacs why was this PR and all related issues closed without merging and with no rationale listed? Is this intended behavior? It seems surprising and unintuitive.

ljharb:
@elashpear this PR was *merged* via landing a squashed commit - see the hash in the closing message.

PR #321, opened 2020-04-10
subset(): test if one range is a subset of another

isaacs:
cc @ljharb


test/ranges/subset.js
```diff
@@ -0,0 +1,48 @@
+const t = require('tap')
+const subset = require('../../ranges/subset.js')
+
+// sub, dom, expect, [options]
+const cases = [
+  ['1.2.3', '1.2.3', true],
+  ['1.2.3 1.2.4', '1.2.3', true],
```
ljharb, line 7: this seems off - the first one allows 1.2.4, but the second one doesn’t, so shouldn’t it be false?

test/ranges/subset.js
```diff
@@ -0,0 +1,48 @@
...
+  ['1.2.3 2.3.4 || 2.3.4', '3', false],
+  ['^1.2.3-pre.0', '1.x', false],
+  ['^1.2.3-pre.0', '1.x', true, { includePrerelease: true }],
+  ['>2 <1', '3', true],
+  ['1 || 2 || 3', '>=1.0.0', true],
+  ['*', '*', true],
+  ['*', '>=0.0.0-0', false], // XXX technically wrong, should detect this?
+  ['>=1.0.0 <=1.0.0 || 2.0.0', '1.0.0 || 2.0.0', true],
+  ['>=1.0.0 <=1.0.0 || 2.0.0', '1.0.0 || 2.0.0', true],
```
ljharb, line 16: this is a repeat of the previous line

test/ranges/subset.js
```diff
@@ -0,0 +1,48 @@
...
+  ['1.2.3', '1.2.3', true],
+  ['1.2.3 1.2.4', '1.2.3', true],
+  ['1.2.3 2.3.4 || 2.3.4', '3', false],
+  ['^1.2.3-pre.0', '1.x', false],
+  ['^1.2.3-pre.0', '1.x', true, { includePrerelease: true }],
+  ['>2 <1', '3', true],
+  ['1 || 2 || 3', '>=1.0.0', true],
+  ['*', '*', true],
+  ['*', '>=0.0.0-0', false], // XXX technically wrong, should detect this?
+  ['>=1.0.0 <=1.0.0 || 2.0.0', '1.0.0 || 2.0.0', true],
```
ljharb, line 15: does the former not allow 3, while the latter does not, hence false?

test/ranges/subset.js
isaacs: The first one is a null set.  It's `1.2.3` AND `1.2.4`, not `1.2.3 || 1.2.4`.

test/ranges/subset.js
isaacs: yep.  fixing.

test/ranges/subset.js
isaacs: The former does not allow 3.  It is:

- versions that are `>=1.0.0` AND `<=1.0.0` (ie, only 1.0.0!), OR
- versions that are 2.0.0.

So the sets are equivalent.

test/ranges/subset.js
ljharb: gotcha

test/ranges/subset.js
ljharb: ah, the above means the former is "1.0.0 || 2.0.0", so true is correct :-)

PR #325, opened 2020-05-06
add support for esm named exports

MylesBorins:
Through an ESM wrapper and package.exports this change allows the
semver module to be both imported and required while still supporting
legacy versions of Node.js. An important advantage to this change is that
semver will now support named imports, which are unsupported in commonjs
modules.

One thing to note. This change introduces the package.exports field
which locks down what modules can be deeply imported.

Currently support is maintained to allow deep imports of all files
currently included in package.files to keep the change Semver Minor.
In the future this interface could be further locked down.


esm-wrapper.mjs
```diff
@@ -0,0 +1,145 @@
...
+const minVersion = require('./ranges/min-version');
+const validRange = require('./ranges/valid');
+const outside = require('./ranges/outside');
+const gtr = require('./ranges/gtr');
+const ltr = require('./ranges/ltr');
+const intersects = require('./ranges/intersects');
+const simplifyRange = require('./ranges/simplify');
+const subset = require('./ranges/subset');
+
+export default {
```
MylesBorins, line 53: unsure if you want to support the default or not... we could remove this if you want to keep the interface limited

esm-wrapper.mjs
```diff
@@ -0,0 +1,145 @@
+// just pre-load all the stuff that index.js lazily exports
+import {createRequire} from 'module';
+
+const require = createRequire(import.meta.url);
+
+const internalRe = require('./internal/re');
```
guybedford, line 6: Please do `import internalRe from './internal/re'` here since this will break bundlers otherwise.

esm-wrapper.mjs
```diff
@@ -0,0 +1,145 @@
...
+
+const re = internalRe.re;
+const src = internalRe.src;
+const tokens = internalRe.t;
+const SEMVER_SPEC_VERSION = require('./internal/constants').SEMVER_SPEC_VERSION;
```
guybedford, line 11: And similarly, `import { SEMVER_SPEC_VERSION } from './internal/constants'` etc.

esm-wrapper.mjs
coreyfarrell: The default export is required for this to not be semver-major, otherwise code that already uses `import semver from 'semver'` will break.

Side note the approach I've taken for semver-minor ESM wrappers is as follows.  @guybedford do you see issues with this for bundlers?
```js
import semver from './index.js'

export const {/* list all named exports here */} = semver

export default semver
```

Then you can test with something like:
```js
// test/esm/test-exports.mjs
import {createRequire} from 'module'
import t from 'tap'

import * as semver from 'semver'

const require = createRequire(import.meta.url)

const cjs = require('semver')

t.equal(cjs, semver.default)
t.same(cjs, Object.fromEntries(Object.entries(semver).filter(nvp => nvp[0] !== 'default')))
```

This verifies that `package.json#exports` is correct for require and import on the main path, then it verifies identical API among the conditional exports.

MylesBorins: So there is an interesting issue with semver... it installs a module which depends on semver, which results in it being in the node_modules folder... which breaks self-referential modules afaict

coreyfarrell: Not when I tested my suggestion locally on node.js 14.1.0, `node_modules/semver` was ignored by the self referential import and require.

MylesBorins:
Oh ok, maybe I was just confused because it wasn't failing as expected on
10.x, glad to hear that is working as expected

isaacs:
Just added a test for `map.js`, which wasn't getting tested right now.

Rebase this onto master, and let the test guide you, and all will be good :)

MylesBorins:
Updated based on all the feedback. Currently have the single test per file but this is now going to fail on all version of node lower than 14 out of the box (for now). Before I was using a proxy cjs module to sniff the semver and skip the esm test on node.js older than 14. How would you suggest to unbreak older versions without compromises the 1 test per file system?

coreyfarrell:
> This LGTM, assuming that it still has full coverage after making the requested modification to the tests.

On it's own nyc will not produce coverage for ES modules loaded by node.js core.  To get that you would need the experimental [@istanbuljs/esm-loader-hook](https://github.com/istanbuljs/esm-loader-hook).  I'm not sure getting coverage reported for the one statement in the esm wrapper is worth the trouble/risk of future breakage by using `--experimental-loader`.

coreyfarrell:
One more comment on ESM coverage, the experimental module I mentioned does not work with nyc 14 (used by the current version of tap).  It's @isaacs call but I think not having automated coverage for the esm wrapper should be fine, being a single statement means that we can see that it's fully covered.

As for the issue with ignoring the test in node.js < 13 I think you could:
* Add `"test-ignore": "test/esm-wrapper.mjs"` to `package.json#tap`
* Add to test/index.js:
```js
if (semver.gte(process.version, '14.0.0')) {
  t.spawn(process.execPath, require.resolve('./esm-wrapper.mjs'))
}
```

Personally I would lean towards renaming the `esm-wrapper.mjs` files to `index.mjs`.  AFAIK the `map.js` settings are applied per test run directly by tap and not to tests run by `t.spawn`.  Doesn't make a difference now but eventually we'll be able to get coverage.  Since `test/index.js` will filter coverage to `index.js` and `index.mjs`, that same filter would be inherited in the t.spawn of test/index.mjs.

package.json
```diff
@@ -21,8 +21,8 @@
     "./package": "./package.json"
   },
   "scripts": {
-    "test": "tap --no-esm",
```
coreyfarrell: I'm confused, why did you remove the `--no-esm`?

MylesBorins: the `t.spawn` shells out to node not tap so it doesn't end up using the built in runner (and esm). At least that is my understanding

isaacs: It'd be better to put `"tap": { "esm": false }` in the package.json than have it hard-coded in the command, imo.

coreyfarrell: Fair enough.  I was just seeing the `esm` module as unnecessary but I guess it's also not strictly required to disable it in this repository.

isaacs:
Yeah, missing coverage for a one-line file is not the end of the world, especially if we can be reasonably sure that it's working properly by some other means.

You can add the test file but include it in `test-ignore` like @coreyfarrell suggests, and then load it on versions that will do the right thing, that's fine.  The nice/hazardous thing about a coverage-map is that if the test isn't run, it's as if its covered file just doesn't exist, so you still get 100%.

MylesBorins:
Are folks happy with what we currently have?

mbruning24:
Currently working on an upgrade to angular 10 for our application, and node-semver is a dependency we have, it's throwing a lot of errors not having named exports, seems like the PR is basically good to go, can we get a bump?

basz:
Not sure this is related, but I'm seeing an error in production builds only. 

"Error: Cannot find module './ranges/outside'" 

When I compare the `node_modules/semver/ranges` directories it is indeed missing an `outside.js` file.

Any way around this?

basz:
nvm. It was an incorrect ignore in ember-electron, filtering all files staring with 'out'...

cawa-93:
Is there any updates? It seems to me that this could help the packages above to get around the [problem of cyclic dependencies](https://github.com/npm/node-semver/issues/381).

MylesBorins:
Closing in lieu of https://github.com/npm/node-semver/pull/383

PR #358, opened 2020-12-30
Add CodeQL security scanning

smorimoto:
This PR adds a static analysis tool called CodeQL which scans your code at build time to find any potential security issues.

PR #359, opened 2020-12-30
Remove useless regular-expression character escape

smorimoto:
This PR fixes the problem found by #358 (though it's hardly a problem).


smorimoto:
@lukekarrys Brilliant!

PR #375, opened 2021-03-18
fix(subset): check any as superset

### LINKED ISSUE #374: [BUG] Invalid falsey response when checking range is subset of any
# What / Why

I've run into an interesting issue when submitting a PR for npm's arborist library, where a range (e.g. `~3.5.0`) is not a subset of any. This should be supported, as even though any is not a subset of a range (which is already handled), the opposite is not the case.

I think the fix would be to add an additional short-circuit check in [lines 58-62 (v7.3.4 snippet)](https://github.com/npm/node-semver/blob/v7.3.4/ranges/subset.js#L58-L62) to handle an any `dom` value:
```diff
  if (sub === dom)
    return true

+  if (dom.length === 1 && dom[0].semver === ANY)
+    return true

  if (sub.length === 1 && sub[0].semver === ANY)
    return dom.length === 1 && dom[0].semver === ANY
```

## When

I run a subset comparison for if a given range is a subset of any

**Example**

```js
const semver = require("semver");
// ~3.5.0 := >=3.5.0 <3.(5+1).0 := >=3.5.0 <3.6.0-0
console.log(semver.subset("~3.5.0", "*"));
// false
```

### Current Behavior

Result is falsey.

### Expected Behavior

Result should be truthy, as `>=3.5.0 <3.6.0-0` should be a subset of any.

## References

* npm/arborist#253 ([this comment](https://github.com/npm/arborist/issues/253#issuecomment-800769581)).

jameschensmith:
## Summary

Adds short-circuit check if superset is `*`. This is currently not handled. A non-range semver has been able to get passed this not existing, but ranges came back falsey. This request addresses that shortcoming.

## References

Fixes #374.

### DIFF
--- a/test/ranges/subset.js
+++ b/test/ranges/subset.js
@@ -15,6 +15,12 @@ const cases = [
   ['>2 <1', '3', true],
   ['1 || 2 || 3', '>=1.0.0', true],
 
+  // everything is a subset of *
+  ['1.2.3', '*', true],
+  ['^1.2.3', '*', true],
+  ['^1.2.3-pre.0', '*', true],
+  ['1 || 2 || 3', '*', true],
+
   ['*', '*', true],
   ['', '*', true],
   ['*', '', true],
--- a/ranges/subset.js
+++ b/ranges/subset.js
@@ -7,6 +7,8 @@ const compare = require('../functions/compare.js')
 // - Every simple range `r1, r2, ...` is a subset of some `R1, R2, ...`
 //
 // Simple range `c1 c2 ...` is a subset of simple range `C1 C2 ...` iff:
+// - If C is only the ANY comparator
+//   - return true
 // - If c is only the ANY comparator
 //   - If C is only the ANY comparator, return true
 //   - Else return false
@@ -56,6 +58,9 @@ const subset = (sub, dom, options) => {
 
 const simpleSubset = (sub, dom, options) => {
   if (sub === dom)
+    return true
+
+  if (dom.length === 1 && dom[0].semver === ANY)
     return true
 
   if (sub.length === 1 && sub[0].semver === ANY)


jameschensmith:
@isaacs, when you get a chance could you take a look at this request? If there's any concerns, I'll do my best to address them ASAP. The request for arborist in the **References** section of this request has a dependency on this being merged. Both of these would also resolve npm/cli#636 (and possibly its referenced tickets as well), so this has potential to resolve a ton of issues for users. Thank you!

test/ranges/subset.js
```diff
@@ -15,6 +15,12 @@ const cases = [
   ['>2 <1', '3', true],
   ['1 || 2 || 3', '>=1.0.0', true],
 
+  // everything is a subset of *
+  ['1.2.3', '*', true],
+  ['^1.2.3', '*', true],
+  ['^1.2.3-pre.0', '*', true],
```
isaacs, line 21: This is incorrect, though.  `^1.2.3-pre.0` is not a subset of `*` unless `includePrerelease` is set, because there are versions that satisfy `^1.2.3-pre.0` and do not satisfy `*`.

```
> semver.satisfies('1.2.3-pre.0', '^1.2.3-pre.0')
true
> semver.satisfies('1.2.3-pre.0', '*')
false
```

isaacs (changes requested):
The suggested short-circuit path can only be applied when the `includePrerelease` flag is set.  Otherwise, it returns an incorrect result if the first range includes a prerelease on any of its versions.

isaacs:
Ah, ok, this is a bug though:

```
> semver.subset('^1.2.3', '*')
false
```

I think we need special handling of `<1.2.3-0`, since that range _doesn't_ include prereleases, maybe?

isaacs:
Ahh!  This was already a problem, turns out!

```
> require('./').subset('^1.2.3-pre.0', '>=1')
true
```

That should return false in non-prerelease mode, because the first set includes `1.2.3-pre.0` and up prereleases, and the second doesn't include any prereleases.

Ok, so the solution looks like this I think:

```
// - If c is only the ANY comparator
//   - If C is only the ANY comparator, return true
//   - Else return false
// - If C is only the ANY comparator
//   - if in prerelease mode, return true
//   - else replace C with `[>=0.0.0]`
// ... rest of the algorithm
// - handle any prerelease ranges in c, ensuring that they don't pull in prereleases not in C
```

isaacs:
This is also a bug:

```
> require('./').subset('*', '>=0.0.0-pre.0')
false
```

Because (in non-prerelease mode) `>=0.0.0-pre.0` includes all versions that could possibly match `*` _as well as_ any versions from `0.0.0-pre.0` through `0.0.1`, it should be a superset of `*`.  So that first bit should be:

```
// - If c is only the ANY comparator
//   - If C is only the ANY comparator, return true
//   - Else if in prerelease mode, return false
//   - else replace c with `[>=0.0.0]`
```

jameschensmith:
Woah! Thanks for your great feedback! I had no idea about the pre-release matching, but it makes sense :blush: I see you set up another PR, requesting my review, so I'll check that out now. Thanks! :+1:

isaacs:
This is landed in #377.

Thanks for bringing up the issue!  Was a pretty significant oversight, which didn't matter until Arborist started making this method load-bearing 😬

PR #377, opened 2021-03-22
Correctly handle prereleases/ANY ranges in subset

isaacs:
An "ANY" range (ie, `""`, `*`, etc.) does not include prerelease
versions except when `includePrerelease` flag is set.

Also, merely looking at the max/min boundaries of any ranges ignores the
fact that the sub range maybe including prerelease versions that are
excluded from the super range.  For example, `>=1.2.3-pre.0` is _not_ a
subset of `>=1.0.0`, because it inludes `1.2.3-pre.0`, `1.2.3-pre.1`,
and so on.

---

@jameschensmith, if you would be so kind as to give this a look over, and make sure it fixes the bug you found, I would greatly appreciate it.  Thanks!

## References

Related to #374, #375 (includes the patch from #375).

### DIFF
--- a/ranges/subset.js
+++ b/ranges/subset.js
@@ -1,22 +1,28 @@
 const Range = require('../classes/range.js')
-const { ANY } = require('../classes/comparator.js')
+const Comparator = require('../classes/comparator.js')
+const { ANY } = Comparator
 const satisfies = require('../functions/satisfies.js')
 const compare = require('../functions/compare.js')
 
 // Complex range `r1 || r2 || ...` is a subset of `R1 || R2 || ...` iff:
-// - Every simple range `r1, r2, ...` is a subset of some `R1, R2, ...`
+// - Every simple range `r1, r2, ...` is a null set, OR
+// - Every simple range `r1, r2, ...` which is not a null set is a subset of
+//   some `R1, R2, ...`
 //
 // Simple range `c1 c2 ...` is a subset of simple range `C1 C2 ...` iff:
-// - If C is only the ANY comparator
-//   - return true
 // - If c is only the ANY comparator
 //   - If C is only the ANY comparator, return true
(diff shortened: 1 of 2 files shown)


ranges/subset.js
```diff
@@ -23,13 +31,16 @@ const compare = require('../functions/compare.js')
 // - If GT
 //   - If GT.semver is lower than any > or >= comp in C, return false
 //   - If GT is >=, and GT.semver does not satisfy every C, return false
+//   - If GT.semver has a prerelease, and not in prerelease mode
+//     - If no C has a prerelease and the GT.semver tuple, return false
 // - If LT
 //   - If LT.semver is greater than any < or <= comp in C, return false
 //   - If LT is <=, and LT.semver does not satisfy every C, return false
-// - If any C is a = range, and GT or LT are set, return false
+//   - If GT.semver has a prerelease, and not in prerelease mode
```
jameschensmith, line 39: I think this should be the following (could be wrong):

```suggestion
//   - If LT.semver has a prerelease, and not in prerelease mode
```

test/ranges/subset.js
```diff
@@ -15,6 +15,29 @@ const cases = [
   ['>2 <1', '3', true],
   ['1 || 2 || 3', '>=1.0.0', true],
 
+  // everything is a subset of *
```
jameschensmith, line 18: This comment should probably say something new, since this is not always true :sweat_smile:

jameschensmith (commented):
Ahh, seems I was a bit too late :see_no_evil: I just had a few minor suggestions, but it looks good, though! Still trying to wrap my head around the range comparisons for pre-releases `includePrerelease`, that's what was taking me a bit :yum: Thanks for addressing this! Much appreciated :blush::+1:

PR #380, opened 2021-05-01
fix: inc prerelease with numeric preid

ntkme:
The [semver spec](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions) says prerelease can be numeric:

```
<pre-release identifier> ::= <alphanumeric identifier>
                           | <numeric identifier>
```

This PR fixes `inc` `prerelease` with numeric `preid`.

E.g. `inc('1.0.0-1.0', 'prerelease', '1')` should return `1.0.0-1.1`, but today it returns `1.0.0-1.0`.

The bug is that currently it compares preid with `this.prerelease[0] === identifier`, and a string `'1'` is not equal to number `1`, thus it's returning incorrect result. Change this comparison to use `compareIdentifiers` just like everywhere else fixes the problem.

- Fixes https://github.com/npm/cli/issues/3181

### DIFF
--- a/classes/semver.js
+++ b/classes/semver.js
@@ -265,7 +265,7 @@ class SemVer {
         if (identifier) {
           // 1.2.0-beta.1 bumps to 1.2.0-beta.2,
           // 1.2.0-beta.fooblz or 1.2.0-beta bumps to 1.2.0-beta.0
-          if (this.prerelease[0] === identifier) {
+          if (compareIdentifiers(this.prerelease[0], identifier) === 0) {
             if (isNaN(this.prerelease[1])) {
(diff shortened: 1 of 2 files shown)


ntkme:
The current version of `node-tap` in this repo won't work on node@16, which is causing CI failure.

ntkme:
@lukekarrys Sorry for pinging but this PR is a one line bug fix that has been sitting for nearly one year without getting any attention, even though it has been confirmed as a bug in https://github.com/npm/cli/issues/3181. Can you please take a look? Thanks.

lukekarrys (approved):

PR #383, opened 2021-05-13
fix: add support for node.js esm auto exports

MylesBorins:
This is an alternative to https://github.com/npm/node-semver/pull/325/

Node.js now supports ESM named exports via auto-detection, but the way `node-semver` was creating `module.exports` was breaking that algorithm. The advantage of this approach is that there is no need for a wrapper file or for `package.exports`. It should be completely Semver Minor

This is a minor tweak to the layout of `index.js` that allows for ESM named exports support in all modern Node.js ESM implementations.

We could borrow the tests from #325, but it it a bit obtuse to add a single test to the testing framework as there is a 1:1 file to test expectation.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,48 +1,88 @@
 // just pre-load all the stuff that index.js lazily exports
 const internalRe = require('./internal/re')
+const constants = require('./internal/constants')
+const SemVer = require('./classes/semver')
+const identifiers = require('./internal/identifiers')
+const parse = require('./functions/parse')
+const valid = require('./functions/valid')
+const clean = require('./functions/clean')
+const inc = require('./functions/inc')
+const diff = require('./functions/diff')
+const major = require('./functions/major')
+const minor = require('./functions/minor')
+const patch = require('./functions/patch')
+const prerelease = require('./functions/prerelease')
+const compare = require('./functions/compare')
+const rcompare = require('./functions/rcompare')
+const compareLoose = require('./functions/compare-loose')
+const compareBuild = require('./functions/compare-build')
+const sort = require('./functions/sort')
+const rsort = require('./functions/rsort')
+const gt = require('./functions/gt')
+const lt = require('./functions/lt')
+const eq = require('./functions/eq')
+const neq = require('./functions/neq')
+const gte = require('./functions/gte')
+const lte = require('./functions/lte')
+const cmp = require('./functions/cmp')
+const coerce = require('./functions/coerce')
+const Comparator = require('./classes/comparator')
+const Range = require('./classes/range')
+const satisfies = require('./functions/satisfies')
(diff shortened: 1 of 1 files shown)


btakita:
Any thoughts on merging this in?

ljharb:
I wonder if perhaps cjs-module-lexer could instead be updated to understand this pattern?

rxliuli:
Is there anything preventing this PR from being merged?

ljharb:
@rxliuli merging this PR would mean that multiple packages in the ecosystem would need to follow it, and old versions never would - whereas https://github.com/npm/node-semver/pull/383#issuecomment-883865501 would mean that *every* version of any package following this package's pattern would Just Work - i don't think one-off changes to packages are a sustainable approach.

MylesBorins:
I still think we should land this @darcyclarke thoughts?

@ljharb I get where you are coming from regarding a philosophical approach here... but this change will in fact fix a usability issue with the package and is ready to land, not landing it on principal seems like a mistake. That doesn't mean we should try and fix this in the lexer as well. /cc @guybedford

guybedford:
@MylesBorins because of backporting constraints, I don't think lexer features should be changed. Object detection is just patchy unfortunately. I think we should only stick to updates for parser bug fixes that break new language syntax or fail on valid language syntax.

ljharb:
@guybedford what would be the backport constraint here? wouldn't this just add new named exports when backported?

guybedford:
The problem is we cannot expect users to track fine-grained exports detection rules as being compatible or not. For example if we made the lexer support these named exports in Node.js 18 only, it would be highly suprising when it breaks on older versions of Node.js, as opposed to other types of new features which can be clearly communicated as experimental.

ljharb:
Understood - although that seems true of virtually any feature, whether labelled experimental or not, and the solution to all of them is to run CI on supported node versions.

guybedford:
Yes, although the difference here is parser internals are difficult to communicate as features, in the same way that versioning parsers or transformers is hard.

index.js
```diff
@@ -1,48 +1,88 @@
 // just pre-load all the stuff that index.js lazily exports
 const internalRe = require('./internal/re')
+const constants = require('./internal/constants')
+const SemVer = require('./classes/semver')
+const indentifiers = require('./internal/identifiers')
```
ljharb, line 5:
```suggestion
const identifiers = require('./internal/identifiers')
```

wraithgar (approved):

wraithgar:
Fixing this here fixes auto esm exports for the most highly downloaded package on the entire npm registry.  Fixing the lexer is outside the scope of what node-semver needs to worry about today, and folks can go discuss that in the appropriate repo.

PR #389, opened 2021-07-01
feat(range): update for safer key

bluelovers:

### DIFF
--- a/classes/range.js
+++ b/classes/range.js
@@ -81,7 +81,7 @@ class Range {
 
     // memoize range parsing for performance.
     // this is a very hot path, and fully deterministic.
-    const memoOpts = Object.keys(this.options).join(',')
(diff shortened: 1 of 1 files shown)


classes/range.js
```diff
@@ -81,7 +81,7 @@ class Range {
 
     // memoize range parsing for performance.
     // this is a very hot path, and fully deterministic.
-    const memoOpts = Object.keys(this.options).join(',')
+    const memoOpts = Object.keys(this.options).filter(k => this.options[k]).join(',')
```
ljharb, line 84: this change invalidates the previous comment by making it not deterministic, no?

lukekarrys: `this.options` is always run through `parse-options` which implements this same filtering, so this change isn't necessary here.

https://github.com/npm/node-semver/blob/cb1ca1d5480a6c07c12ac31ba5f2071ed530c4ed/internal/parse-options.js#L7

PR #411, opened 2021-10-26
don't call the second number a major

kokushkin:
major is always the first number, calling the second a major might be a very confusing. Took me a lot of time to recover from this false belief.

## References

https://github.com/raineorshine/npm-check-updates/issues/958


Ahmedelsayedfarraj (approved):

wraithgar:
The docs as is accurately describe the nuances with leading zeroes in caret ranges.
Removing these sections adds more confusion because it is important to call out why this behavior exists and why one would want to use caret ranges for versions that have leading zeroes.

PR #420, opened 2022-01-15
docs: use "semver" instead of "SemVer" where appropriate

rauno56:
Noticed that "semver" and "SemVer" is being used interchangeably in the docs and bin help output. Made that consistent.


wraithgar:
Looking a little more closely at this, it appears to be an intentional distinction between the package name (npm packages are all lower case) and the SemVer spec itself, which is capitalized.

PR #401, opened 2021-09-18
update tap to latest to run test locally

sify21:
Signed-off-by: 司芳源 <sify1221@gmail.com>

I can not run tests locally using node version `v16.8.0`. 
- changed tap's version specifier to allow using latest version.
- changed `t.done()` to `t.end()` in `test/functions/coerce.js`

There are still 3 failed tests, I'm not familiar with Promise and don't know how to fix them.
```
 FAIL  test/bin/semver.js 3 failed of 4 821.243ms
 ✖ child test left in queue: t.test help output
 ✖ child test left in queue: t.test sorting and filtering
 ✖ child test left in queue: t.test coercing
```

## References
Related to https://github.com/tapjs/node-tap/issues/746

PR #426, opened 2022-01-28
test: update tap + fix tests

louis-bompart:
I had the following error when running `npm test`
```
> semver@7.3.5 test
> tap

TypeError: Cannot read properties of undefined (reading 'getFileName')
    at module.exports (node-semver\node_modules\tap\node_modules\caller-path\index.js:4:40) 
    at importJSX (node-semver\node_modules\tap\node_modules\import-jsx\index.js:24:46)      
    at module.exports (node-semver\node_modules\tap\node_modules\treport\lib\index.js:13:18)
    at exports.makeReporter (node-semver\node_modules\tap\bin\run.js:422:23)
    at runTests (node-semver\node_modules\tap\bin\run.js:744:3)
    at mainAsync (node-semver\node_modules\tap\bin\run.js:244:5)
TAP version 13
1..0
# time=1.581ms
```

Updating `tap` allowed the to run but failed on `test/functions/coerce.js`:
```
 FAIL ​ test/functions/coerce.js 1 failed of 93 43.195ms
 ✖ t.done is not a function
```
 
I replaced `t.done` with `t.end`, which, as far as I can tell, is used everywhere else.

The test run was then fully green.

## References

 Related to #401 - Actually the same, just with a more recent version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "semver",
       "version": "7.3.5",
       "license": "ISC",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


lukekarrys:
Closing in favor of #433

PR #427, opened 2022-01-28
fix: ensure SemVer instance passed to inc are not modified

### LINKED ISSUE #425: [BUG] `inc` affects the inputed SemVer object.
# What / Why

> `inc` when called with an instance of `SemVer` will affect the provided `SemVer` object (if no opts are given, see Steps to Reproduce). This leads to unexpected behaviour for a function that could be expected to be pure.

## When

* n/a

## Where

* n/a

### Current Behavior

`inc` will do the operation on the SemVer object provided in parameters, except if semver is created with `loose` or `includePrereleaseSemVer`.

### Steps to Reproduce

```js
const { SemVer, inc } = require("semver");
const { equal } = require("assert");

const initalStrVer = "1.0.0";
const initialSemver = new SemVer(initalStrVer);
const looseSemver = new SemVer(initalStrVer, { loose: true });
const includePrereleaseSemVer = new SemVer(initalStrVer, {
  includePrerelease: true,
});

inc(initialSemver, "minor");
inc(looseSemver, "minor");
inc(includePrereleaseSemVer, "minor");
inc(initalStrVer, "minor");

equal(initalStrVer, looseSemver.version); // ✔️  Pass
equal(initalStrVer, includePrereleaseSemVer.version); // ✔️ Pass
equal(initalStrVer, initialSemver.version); // ❌ AssertionError [ERR_ASSERTION]: '1.0.0' == '1.1.0'
```

### Expected Behavior

`inc` should never modify the provided SemVer object.

## Who

* n/a

## References

  * Related to #378

louis-bompart:
`inc` can be used three ways:
 1.  As a an option on the CLI : `npx semver 1.0.0 -i major`
 2.  As a method of the `SemVer` class: `semverInstance.inc('major')`
 3.  As a function: `inc('1.0.0', 'major')`
 
 This PR does not impact the behaviour of the first two and focuses only on the last one.
 
 The underlying issue is described in #425, but tl;dr `inc` should be a pure function (I think, from its documentation on NPM), and here it has a side-effect.
 
 I remove this side-effect by ensuring that the `SemVer` instance created inside the `inc` function is created from a string, which removes the mutation vector.

## References
  Depends on #426 - just to fix the tests, that's why https://github.com/npm/node-semver/commit/563d75e3b5c10ee72c3d05c38c55c8a0e6c83c98 is polluting the changeset
  Fixes #425

### DIFF
--- a/functions/inc.js
+++ b/functions/inc.js
@@ -7,7 +7,10 @@ const inc = (version, release, options, identifier) => {
   }
 
   try {
-    return new SemVer(version, options).inc(release, identifier).version
+    return new SemVer(
+      version instanceof SemVer ? version.version : version,
+      options
+    ).inc(release, identifier).version
   } catch (er) {
     return null
   }
--- a/test/functions/inc.js
+++ b/test/functions/inc.js
@@ -10,10 +10,20 @@ test('increment versions test', (t) => {
     t.equal(found, wanted, `${cmd} === ${wanted}`)
 
     const parsed = parse(pre, options)
+    const parsedAsInput = parse(pre, options)
     if (wanted) {
       parsed.inc(what, id)
       t.equal(parsed.version, wanted, `${cmd} object version updated`)
       t.equal(parsed.raw, wanted, `${cmd} object raw field updated`)
+
+      const preIncObject = JSON.stringify(parsedAsInput)
+      inc(parsedAsInput, what, options, id)
+      const postIncObject = JSON.stringify(parsedAsInput)
+      t.equal(
+        postIncObject,
+        preIncObject,
+        `${cmd} didn't modify its input`
+      )
     } else if (parsed) {
       t.throws(() => {
         parsed.inc(what, id)


lukekarrys:
Thanks for this PR @louis-bompart! I think this is the right direction for `inc` as well as any other "pure" functions. Would you be willing to update this PR with this change made across all those functions? I also think we could export `semver/functions/clone` that would clone an instance and use that internally.

And I apologize for letting this sit so long, it now has some conficts and shouldn't need any dep updates as those have all been updated.

lukekarrys (changes requested):
See previous comment

louis-bompart:
> I also think we could export `semver/functions/clone` that would clone an instance and use that internally.

That's my two cents, so take it as it is, but I think that a 'clone' function would be a bit out of place:
As a user, I'd be confused about what 'cloning' a semver is doing, especially when the other functions are quite so 'math-oriented'.

I think adding a factory function to the SemVer class would maybe be more in place.

So instead of doing:
```js
import {clone, SemVer} from 'semver';
const verA = new SemVer('1.0.0');
const verB = clone(verA);
```

One would do this:
```js
import {SemVer} from 'semver';
const verA = new SemVer('1.0.0');
const verB = SemVer.clone(verA);
```

Another solution would be to do nothing and use [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) internally. This would require to enforce Node >=17.x tho.

louis-bompart:
> Would you be willing to update this PR with this change made across all those functions?
✔️. As far as I could tell, only `parse` was returning an actual SemVer object from a SemVer object.

lukekarrys:
Thanks @louis-bompart! This looks great!

> As far as I could tell, only parse was returning an actual SemVer object from a SemVer object.

Thanks for looking into this, not sure why I thought there would be more. And sorry for the back and forth but I now think that this should only apply to `inc` since that is designed to change the state of a version.

I can see people using `parse` as a safe way to handle a variable that might be a string or a semver instance and I don't want to break the strict equality that returning the instance provides.

```js
> s = new semver.SemVer('1.0.0')
SemVer {
  options: {},
  loose: false,
  includePrerelease: false,
  raw: '1.0.0',
  major: 1,
  minor: 0,
  patch: 0,
  prerelease: [],
  build: [],
  version: '1.0.0'
}
> s === semver.parse(s)
true
> s === semver.parse('1.0.0')
false
```

lukekarrys (changes requested):
Can you remove the commit updating `parse`? After that I think this is good to go!

lukekarrys (approved):

PR #432, opened 2022-02-26
fix: properly escape dots in `GTE0` regexes

lukekarrys:
Previously the dots were not properly escaped causing them to match any
character. This caused ranges like `>=09090` to evaulate to `*` after
they were incorrectly matched by the `GTE0` regex. This only happened in
strict mode since in loose mode the leading 0 is allowed and parsed into
`>=9090.0.0` before the `GTE0` check. After this fix, this range now
will throw an error. This also affected prerelease versions in both
strict and loose mode.


lukekarrys:
This also includes a small amount of refactoring and a few added debug statements that made it easier for me to dig into the details here. The behavior and tests were trivial to change, but it was my first time in the codebase so I was curious to see why it happened.

nlf (approved):
this looks good!

PR #433, opened 2022-02-26
chore: use `@npmcli/template-oss`

### LINKED ISSUE #146: Standard?

### LINKED ISSUE #421: [FEATURE] Changelog
There is no changelog for versions > 7.3.0, neither in the CHANGELOG.md nor in the tag descriptions.

I would appreciate it if you could outline the changes so consupers of node-semver don’t have to crawl through the commit history to see if it is safe to upgrade.

lukekarrys:
This adds `@npmcli/template-oss` to manage GitHub Actions, linting, and
other chores.

This surfaced a few bugs which I opted to fix in separate issues:
- #432
- #434

Closes #146
Closes #421
Closes #426
Closes #401 
Closes #358


lukekarrys:
This fails linting and CodeQL which should be fixed by #432 and #434. I think this should be merged first before those are considered since this will allow for CI to run properly on those PRs.

ljharb:
This was a breaking change, since it dropped a number of node versions that are included in `>= 10` but not the latest semver range; see #447.

PR #434, opened 2022-02-26
fix: replace regex used to split ranges

lukekarrys:
Found in #433, the regex used to split ranges on `||` was padded by `\s*`
on either side causing a decrease in performance when used on range
strings with a lot of spaces. Since the result of the split is
immediately trimmed, we can just split on the string instead.


nlf (approved):
this is very clearly correct :+1:

PR #439, opened 2022-03-11
tap@16.0.0

isaacs:
Fix tests and snapshots to work with new tap version

cc: @wesleytodd


wesleytodd (approved):
Awesome. I think I had some different tests fail, but maybe it was just because I had not updated the snapshots yet.

PR #449, opened 2022-04-10
fix(bin): get correct value from arg separated by equals

### LINKED ISSUE #431: [BUG] bin doesnt work when incrementing arg with equals sign
# What / Why
Why it unshift an empty string into argv when an argument contains a "="?

```javascript
// line 35 @ /bin/semver.js
let a = argv.shift()
const indexOfEqualSign = a.indexOf('=')
if (indexOfEqualSign !== -1) {
  a = a.slice(0, indexOfEqualSign)
  argv.unshift(a.slice(indexOfEqualSign + 1))
}
```

Let's suppose `a` contains an "=", then `indexOfEqualSign` won't be -1. Inside the if statement, it overwrites `a` with a slice of it (removing whatever is after the "="). After that, it inserts into argv another slice of `a`. But this slice starts at `indexOfEqualSign` + 1 (the part we just remove), so an empty string ("") will be returned by the slice and inserted by the unshift.

As I understood, the empty string doesn't affect the execution because it is filtered out at line 92. But, Why it bother to add an "", loop with it, add it to versions to finally remove from versions? I think I am missing something. Can someone explain it to me, please?

lukekarrys:
Fixes #431

### DIFF
--- a/bin/semver.js
+++ b/bin/semver.js
@@ -37,8 +37,9 @@ const main = () => {
     let a = argv.shift()
     const indexOfEqualSign = a.indexOf('=')
     if (indexOfEqualSign !== -1) {
+      const value = a.slice(indexOfEqualSign + 1)
       a = a.slice(0, indexOfEqualSign)
-      argv.unshift(a.slice(indexOfEqualSign + 1))
+      argv.unshift(value)
     }
     switch (a) {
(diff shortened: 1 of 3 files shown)


ljharb (approved):
seems good

Issue #376, opened 2021-03-19
[BUG] Version with v-prefix is incorrectly marked as valid


LouisMT:
According to SemVer 2.0.0, a version like `v1.2.3` is not valid (https://semver.org/#is-v123-a-semantic-version).

The library incorrectly marks a version like this as valid:

```js
const semverValid = require('semver/functions/valid')

semverValid('v1.2.3')
// true

semverValid('v1.2.3', {loose: false})
// true
```

I understand that the v-prefix is often used to indicate that some string represents a version. However, when I explicitly want to check if a string is valid according to the SemVer spec, I expect the method to return false in this case, especially when I set the `loose` option to `false`.

A fix (or an option to be extra strict for backwards compatibility) would be appreciated!

lukekarrys:
As a "fix" this would be a breaking change, so I would be against that. But I do think there is room for a feature that would allow for a more strict parsing according to the spec.

ljharb:
The npm ecosystem doesn’t follow semver 2.0 - it follows 1.0, i believe - because version changes under v1 aren’t always breaking.

That said, i don’t think the v has ever been included according to the spec, but also, in the npm ecosystem it’s very often there. I don’t think there’d be any value in rejecting these.

riderx:
@ljharb that not what is writed in the readme:
[`A "version" is described by the v2.0.0 specification found at https://semver.org/.`](https://github.com/npm/node-semver#versions)
As @LouisMT i expected the validation to return false.
Especially when that is written in the readme one line after `A leading "=" or "v" character is stripped off and ignored.`

ljharb:
Then i guess the docs are wrong.

riderx:
@ljharb that what you believe, to me this package follow `v2.0.0`
According to this [comment](https://github.com/semver/semver/issues/231#issuecomment-104478395) the npm ecosystem does too.
At least for Angular, react, node, typescript, svelte, nuxt, next, and most of the bigger packages.
it's seemed to me fairly legit to include this check in the validation by default in the next MAJOR version, and optional in the current one.

ljharb:
> That said, i don’t think the v has ever been included according to the spec, but also, in the npm ecosystem it’s very often there. I don’t think there’d be any value in rejecting these.

riderx:
I see value in rejecting thoses.
Simple example:
I check SemVer in node ecosystem after these versions are uses in mobile (swift and android) where SemVer is implemented correctly and get refused.
[...]

PR #476, opened 2022-09-06
doc: add warning in readme

riderx:
Following this chat, https://github.com/npm/node-semver/issues/376
This adds to the README to prevent misleading usage of this package, for people like me who expect SemVer package follow the RCF and specifically this : https://semver.org/#is-v123-a-semantic-version

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 semver(1) -- The semantic versioner for npm
 ===========================================
 
+> It is a special node implementation for NPM the RFC is not guarantee
+> 
+> As exemple leading `v` is consider as valid by `valid` method
+
 ## Install
 
 ```bash


lukekarrys:
I agree there is work to do around explicitly documenting differences from the spec for this package. But this PR doesn't create enough clarity to be merged. I'm going to close this but keep #376 open with the `spec` label, so when it does get documented that issue can be found.

riderx:
Thanks for your answer, I understand your point that the PR doesn't add a lot of clarity.
For me that a start, currently, the package is totally not explicit.
Having MVP documentation seems understandable as user.
It will be a first step in the right direction.
Then it will reduce frustration for users and repetitive issue for maintainer.
What do you think?

PR #477, opened 2022-09-07
docs: update range.js comments to clarify the caret ranges examples

amitse:
Update range.js comments to clarify the caret examples

Caret (^) - does not modify the left-most non-zero element in the M.m.p-e. Tilde (~) - does not modify M and m.

hence, ^0.0.1-1 matches 0.0.1-5
however, ~0.0.1-1 matches 0.0.5
which might be counterintuitive and hence adding this example explicitly.


wraithgar:
Thanks, the "leading zeroes with caret ranges" is definitely one of the finer points of semver that trip people up. Anything that adds clarification is probably a good idea.

PR #487, opened 2022-10-04
chore(ci): dont set workspaces flag for tests

lukekarrys:


lukekarrys:
A better fix for this should be in template-oss, but this works for now.

PR #504, opened 2022-11-20
refactor: optimize code

budarin:
Replace suboptimal code with efficient one

### DIFF
--- a/classes/range.js
+++ b/classes/range.js
@@ -104,7 +104,7 @@ class Range {
     range = range.replace(re[t.CARETTRIM], caretTrimReplace)
 
     // normalize spaces
-    range = range.split(/\s+/).join(' ')
+    range = range.replace(/  +/g, ' ')
 
     // At this point, the range is completely trimmed and
     // ready to be split into comparators.
--- a/ranges/to-comparators.js
+++ b/ranges/to-comparators.js
@@ -3,6 +3,6 @@ const Range = require('../classes/range')
 // Mostly just for testing and legacy API reasons
 const toComparators = (range, options) =>
   new Range(range, options).set
-    .map(comp => comp.map(c => c.value).join(' ').trim().split(' '))
+    .map(comp => comp.map(c => c.value.trim()))
 
 module.exports = toComparators


ranges/to-comparators.js
```diff
@@ -3,6 +3,6 @@ const Range = require('../classes/range')
 // Mostly just for testing and legacy API reasons
 const toComparators = (range, options) =>
   new Range(range, options).set
-    .map(comp => comp.map(c => c.value).join(' ').trim().split(' '))
+    .map(comp => comp.map(c => c.value.trim()))
```
ljharb, line 6: this isn't exactly the same thing - `[' a ', ' b ']` would have previously produced `['a ', ' b']` but will now produce `['a', 'b']`.

classes/range.js
```diff
@@ -104,7 +104,7 @@ class Range {
     range = range.replace(re[t.CARETTRIM], caretTrimReplace)
 
     // normalize spaces
-    range = range.split(/\s+/).join(' ')
+    range = range.replace(/  +/g, ' ')
```
ljharb, line 107: this change means it will no longer select non-space whitespace characters, including tabs and other unicode whitespace chars.

ranges/to-comparators.js
budarin: all tests are passed

classes/range.js
budarin: you should provide tests to protect it

ljharb: Indeed, there should be tests, but few projects are willing to change behavior just because there’s not a test yet.

budarin:
There are several similar fragments in the code - they basically solve the task assigned to them, but from the point of view of performance, this is just the worst option - we temporarily create 2 arrays and 2 rows and then throw out their garbage. 

For execution in the browser - this is a window when it comes to working in a server environment where thousands of requests are executed in parallel - a useless load is created for the garbage collector, which negatively affects the performance of the application

lukekarrys (commented):
@budarin can you provide benchmarks showing the performance difference of these changes?

budarin:
There are no benchmarks, but according to the experience of developing services at high loads, the generation of thousands of requests for such garbage turns into jumps in performance - first a pile of such garbage grows, then the process is suspended by the garbage collector and then everything repeats again

Here, even benchmarks are not needed to understand that extra garbage (generating multiple arrays and strings) just to get another string from the source string will lead to memory costs and, accordingly, increase the time for garbage collection

budarin:
Consider storing versions internally not in a string representation

https://marvinh.dev/blog/speeding-up-javascript-ecosystem/
please look at the paragraph "The curious case of semver"

wraithgar:
Closing in favor of the discussion/improvements in https://github.com/npm/node-semver/pull/528

PR #529, opened 2023-04-04
docs: use "semver" instead of "SemVer" where appropriate

wraithgar:
fixup of #420


wraithgar:
Looking a little more closely at this, it appears to be an intentional distinction between the package name (npm packages are all lower case) and the SemVer spec itself, which is capitalized.

PR #249, opened 2018-07-22
Added feature prerelease identifier numbering option - zero or one-based.

b-bly:
Issue link: https://github.com/npm/node-semver/issues/245

PR #532, opened 2023-04-05
feat: identifierBase parameter for .inc

wraithgar:
Rebase and cleanup of #249


nlf (approved):
this makes a lot of sense too, lgtm

PR #334, opened 2020-07-14
fix: semver.diff prerelease to release recognition

deefactorial:
# What / Why
This Bug Fix fixes issues #333 and #288 
This bug is when semver.diff incorrectly identifies releases from a prerelease to patch, minor, or major

- [x] tests included

## References
#333 #288

PR #533, opened 2023-04-05
fix: semver.diff prerelease to release recognition

### LINKED ISSUE #333: [BUG] semver.diff returns incorrect responses
# What / Why

```js
const semver = require('semver');

let releaseArray = ['prepatch', 'prerelease', 'patch', 'preminor', 'prerelease', 'minor', 'premajor', 'prerelease', 'major'];
let v = '0.0.1';

releaseArray.forEach((release) => {
    let bump = semver.inc(v, release)
    let diff = semver.diff(v, bump)

    console.log(`before: '${v}'`, `after: '${bump}'`, `release type: '${release}'`, `diff response: '${diff}'`, `is diff response equal? '${release === diff}'`);
    v = bump;
});
```
console log output: (with spaces added for readability)
```
before: '0.0.1'    after: '0.0.2-0'     release type: 'prepatch'   diff response: 'prepatch'   is diff response equal? 'true'
before: '0.0.2-0'  after: '0.0.2-1'     release type: 'prerelease' diff response: 'prerelease' is diff response equal? 'true'
before: '0.0.2-1'  after: '0.0.2'       release type: 'patch'      diff response: 'prerelease' is diff response equal? 'false'
before: '0.0.2'    after: '0.1.0-0'     release type: 'preminor'   diff response: 'preminor'   is diff response equal? 'true'
before: '0.1.0-0'  after: '0.1.0-1'     release type: 'prerelease' diff response: 'prerelease' is diff response equal? 'true'
before: '0.1.0-1'  after: '0.1.0'       release type: 'minor'      diff response: 'prerelease' is diff response equal? 'false'
before: '0.1.0'    after: '1.0.0-0'     release type: 'premajor'   diff response: 'premajor'   is diff response equal? 'true'
before: '1.0.0-0'  after: '1.0.0-1'     release type: 'prerelease' diff response: 'prerelease' is diff response equal? 'true'
before: '1.0.0-1'  after: '1.0.0'       release type: 'major'      diff response: 'prerelease' is diff response equal? 'false'
```

## When
When I do go from a prerelease to a release (patch, minor, or major) semver.diff returns incorrect responses.

## Where
semver.diff

## How
semver.diff('0.0.2-1', '0.0.2')

### Current Behavior
semver.diff('0.0.2-1', '0.0.2') === 'prerelease'
semver.diff('0.1.0-1', '0.1.0') === 'prerelease'
semver.diff('1.0.0-1', '1.0.0') === 'prerelease'

### Steps to Reproduce
semver.diff('0.0.2-1', '0.0.2')

### Expected Behavior
semver.diff('0.0.2-1', '0.0.2') === 'patch'
semver.diff('0.1.0-1', '0.1.0') === 'minor'
semver.diff('1.0.0-1', '1.0.0') === 'major'

## Who
@deefactorial

## References
#288

wraithgar:
Cleanup of #334 
Closes #333

Authored by @deefactorial

### DIFF
--- a/functions/diff.js
+++ b/functions/diff.js
@@ -10,11 +10,27 @@ const diff = (version1, version2) => {
     const hasPre = v1.prerelease.length || v2.prerelease.length
     const prefix = hasPre ? 'pre' : ''
     const defaultResult = hasPre ? 'prerelease' : ''
-    for (const key in v1) {
-      if (key === 'major' || key === 'minor' || key === 'patch') {
-        if (v1[key] !== v2[key]) {
-          return prefix + key
-        }
+
+    if (v1.major !== v2.major) {
+      return prefix + 'major'
+    }
+    if (v1.minor !== v2.minor) {
+      return prefix + 'minor'
+    }
+
+    if (v1.patch !== v2.patch) {
+      return prefix + 'patch'
+    }
+
+    if (!v1.prerelease.length || !v2.prerelease.length) {
+      if (v1.patch) {
+        return 'patch'
+      }
+      if (v1.minor) {
+        return 'minor'
+      }
+      if (v1.major) {
+        return 'major'
       }
     }
     return defaultResult // may be undefined
--- a/test/functions/diff.js
+++ b/test/functions/diff.js
@@ -4,19 +4,25 @@ const diff = require('../../functions/diff')
 test('diff versions test', (t) => {
 //  [version1, version2, result]
 //  diff(version1, version2) -> result
-  [['1.2.3', '0.2.3', 'major'],
+  [
+    ['1.2.3', '0.2.3', 'major'],
+    ['0.2.3', '1.2.3', 'major'],
     ['1.4.5', '0.2.3', 'major'],
     ['1.2.3', '2.0.0-pre', 'premajor'],
+    ['2.0.0-pre', '1.2.3', 'premajor'],
     ['1.2.3', '1.3.3', 'minor'],
     ['1.0.1', '1.1.0-pre', 'preminor'],
     ['1.2.3', '1.2.4', 'patch'],
     ['1.2.3', '1.2.4-pre', 'prepatch'],
-    ['0.0.1', '0.0.1-pre', 'prerelease'],
-    ['0.0.1', '0.0.1-pre-2', 'prerelease'],
-    ['1.1.0', '1.1.0-pre', 'prerelease'],
+    ['0.0.1', '0.0.1-pre', 'patch'],
+    ['0.0.1', '0.0.1-pre-2', 'patch'],
+    ['1.1.0', '1.1.0-pre', 'minor'],
     ['1.1.0-pre-1', '1.1.0-pre-2', 'prerelease'],
     ['1.0.0', '1.0.0', null],
-
+    ['0.0.2-1', '0.0.2', 'patch'],
+    ['0.1.0-1', '0.1.0', 'minor'],
+    ['1.0.0-1', '1.0.0', 'major'],
+    ['0.0.0-1', '0.0.0', 'prerelease'],
   ].forEach((v) => {
     const version1 = v[0]
     const version2 = v[1]


functions/diff.js
```diff
@@ -17,6 +17,19 @@ const diff = (version1, version2) => {
         }
       }
     }
+    if (v1.major === v2.major &&
+        v1.minor === v2.minor &&
+        v1.patch === v2.patch && !(v1.prerelease.length && v2.prerelease.length)) {
```
H4ad, line 22: Based on line 15 and line 12, we can simplify this to:

```suggestion
if (!hasPre) {
```

H4ad, line 22: Based on the assumptions of line 15, we can reduce this to:

```suggestion
if(!(v1.prerelease.length && v2.prerelease.length)) {
```

functions/diff.js
```diff
@@ -17,6 +17,19 @@ const diff = (version1, version2) => {
...
+      const releaseTypes = []
+      for (const key in v1) {
+        if (key === 'major' || key === 'minor' || key === 'patch') {
+          if (v1[key]) {
+            releaseTypes.push(key)
+          }
+        }
+      }
+      return releaseTypes.pop()
```
H4ad, line 31: Is a minor performance change but if we do like this:

```suggestion
      const key = (v1.major && 0x001) | (v1.minor && 0x010) | (v1.patch && 0x100)

      switch (key) {
        case 0x001:
          return 'patch'

        case 0x010:
          return 'minor'

        case 0x100:
          return 'major'

        case 0x011:
          return 'patch'
        
        case 0x101:
          return 'patch'

        case 0x110:
          return 'minor'

        default:
          return undefined
      }
```

We can go from:

```
diff(0.0.1, 0.0.1-pre) x 303,099 ops/sec ±1.21% (91 runs sampled)
diff(0.0.1, 0.0.1-pre-2) x 305,103 ops/sec ±1.16% (93 runs sampled)
diff(1.1.0, 1.1.0-pre) x 306,851 ops/sec ±0.87% (94 runs sampled)
```

To:

```
diff(0.0.1, 0.0.1-pre) x 318,224 ops/sec ±1.12% (92 runs sampled)
diff(0.0.1, 0.0.1-pre-2) x 320,145 ops/sec ±0.70% (94 runs sampled)
diff(1.1.0, 1.1.0-pre) x 318,362 ops/sec ±0.94% (93 runs sampled)
```

Is a little performance improvement but is something :)

H4ad:
In this line:

https://github.com/npm/node-semver/blob/b3a7d63292dcb7abcab3cb681e80c47061414c4c/functions/diff.js#L8-L9

If we parse it first, we can speedup this from 300k op/s to ~470k~500k op/s.

functions/diff.js
wraithgar: I feel like we can get the same increase by using three `if` statements w/ early return. And it would be much more readable.

H4ad: We can change to a `map` or `hash table`, but a bunch of ifs also works.

If you change these lines to use if, also change the first ones for 3 ifs:

```js
    if (v1.major !== v2.major)
      return prefix + major

    if (v1.minor !== v2.minor)
      return prefix + minor

    if (v1.patch !== v2.patch)
      return prefix + patch
```

functions/diff.js
wraithgar: boolean algebra gives us

`if (!v1.prerelease.length || !v2.prerelease.length) {`

functions/diff.js
wraithgar: I'll update this PR but leave the changes in #539 out.

nlf (approved):
this one was a bit confusing and is a little unintuitive, but it does make sense. for future readers here's what helped me wrap my head around it:

the versions `1.0.0`, `1.1.0-pre` and `1.1.0` would sort in that same order. the transition from `1.0.0` to `1.1.0-pre` _could_ be considered a pre-release, but it's also a minor and since the minor is the more impactful change that's what we report the diff as.

Issue #349, opened 2020-11-11
[BUG] inc() may produce invalid semver with ReleaseType: prerelease


boxcee:
# What / Why
Given an identifier I want to increment a prerelease version.

### Current Behavior
Given an 'invalid' identifier, `inc()` will produce an invalid `semver`.

### Steps to Reproduce
```
const identifier = 'hotfix/410';
const incremented = inc(parse('0.0.1'), 'prerelease', identifier);
console.log(!!valid(incremented));
// => false
```
### Expected Behavior
The documentation of `inc()` specifies this:
```
Return the version incremented by the release type (major, premajor, minor, preminor, patch, prepatch, or prerelease), or null if it's not valid
```
In the case above `inc()` is not returning null, even though the resulting version is not a valid `semver.` Either the documentation needs to be changed to 'or null if the ReleaseType is not valid' or `inc()` needs to check the validity of the `semver` before returning it.

PR #351, opened 2020-11-11
feat: Add function to validate an identifier

boxcee:
## References
Closes #350

PR #534, opened 2023-04-05
feat: Add function to validate an identifier

### LINKED ISSUE #350: [FEATURE] Function to check for validity of a prerelease identifier
# What / Why
Problem described in #349: An 'invalid' identifier will lead to an invalid `semver`.
Now, there is no way to check beforehand whether an identifier is valid, except doing this:
```
const isValidIdentifier(identifier) => !!valid(inc(parse('0.0.1'), 'prerelease', identifier));
```
This is a request for a more straight-forward function, e.g. `identifier()`, which returns `null` (or `false`) if the identifier is invalid.

## References
  * Related to #349

wraithgar:
Cleanup and linting of #351 
Closes #350 

Authored by @boxcee

### DIFF
--- a/functions/identifier.js
+++ b/functions/identifier.js
@@ -1,0 +1,10 @@
+const { t, re } = require('../internal/re')
+const identifier = (id) => {
+  const withHyphen = `-${id}`
+  const match = withHyphen.match(re[t.PRERELEASE])
+  if (match && match[0] === withHyphen) {
+    return id
+  }
+  return null
+}
+module.exports = identifier
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ const semverCompareLoose = require('semver/functions/compare-loose')
 const semverCompareBuild = require('semver/functions/compare-build')
 const semverSort = require('semver/functions/sort')
 const semverRsort = require('semver/functions/rsort')
+const semverIdentifier = require('semver/functions/identifier')
 
 // low-level comparators between versions
 const semverGt = require('semver/functions/gt')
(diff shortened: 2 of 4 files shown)


functions/identifier.js
```diff
@@ -0,0 +1,10 @@
+const { t, src } = require('../internal/re')
+const identifier = (version) => {
+  const split = version.split('.')
+  const valid = split.every(part => new RegExp(`^${src[t.PRERELEASEIDENTIFIER]}$`).test(part))
```
H4ad, line 4: For performance reasons, is better to cache this RegExp instead of creating an instance every time.

wraithgar: I agree. This should exist in `internal/re.js`.

wraithgar:
the codebase calls this a `prereleaseidentifier` which is quite a mouthful

ljharb:
The bin (which is user-exposed, unlike the codebase) calls it `--preid` - perhaps `preID` as short for `preleaseIdentifier`?

wraithgar:
Looking a little deeper at this, the impetus was https://github.com/npm/node-semver/issues/349 which this PR doesn't address.  The line we've been using on these PRs is that if they were ready to go we could land them, but if they needed more work we'd close them and hope that folks who wanted the feature would revisit the work.  I'm gonna close this in light of #349 which is the real bug to be fixed here.

ljharb:
@wraithgar i have a local branch ready to fix #349 once this lands :-/

wraithgar:
Can you implement it without having to export a new function?

ljharb:
Certainly I can, but then consumers won't have an easily accessible predicate function to help explain why they got `null`. What's the argument against adding this?

boxcee:
As I understand the closing comment, this PR needs more work. I cannot see any requests to change anything though, except perhaps the naming.

wraithgar:
In rebasing old PRs the line was drawn at "is this ready" and since it didn't address the underlying issue and folks wanted to bikeshed the name it was deemed "not ready" and closed.

If there were someone actively making a new PR and responding to feedback it would likely be landed.

PR #535, opened 2023-04-06
fix: faster parse options

H4ad:
The performance of parse-options before:

```
includePrerelease x 12,935,785 ops/sec ±3.66% (80 runs sampled)
includePrerelease + loose x 6,915,009 ops/sec ±1.54% (85 runs sampled)
includePrerelease + loose + rtl x 7,350,781 ops/sec ±1.54% (88 runs sampled)
```

After:

```
includePrerelease x 1,122,022,296 ops/sec ±0.81% (96 runs sampled)
includePrerelease + loose x 1,133,747,998 ops/sec ±0.11% (89 runs sampled)
includePrerelease + loose + rtl x 1,135,218,097 ops/sec ±0.35% (95 runs sampled)
```

The performance improvement also extend for any method that use `parseOptions`, like `satisfies`:

Before:

```
satisfies(1.0.6, 1.0.3||^2.0.0) x 281,862 ops/sec ±2.84% (89 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0) x 312,593 ops/sec ±0.90% (95 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0) x 323,066 ops/sec ±1.15% (89 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true}) x 275,542 ops/sec ±1.24% (91 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true}) x 295,613 ops/sec ±1.07% (94 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true}) x 314,728 ops/sec ±1.22% (90 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true}) x 253,795 ops/sec ±0.93% (94 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true}) x 268,212 ops/sec ±1.18% (95 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true}) x 280,362 ops/sec ±1.18% (94 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 218,464 ops/sec ±2.21% (90 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 232,007 ops/sec ±1.05% (91 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 244,718 ops/sec ±0.90% (94 runs sampled)
```

Then:

```
satisfies(1.0.6, 1.0.3||^2.0.0) x 323,247 ops/sec ±1.44% (90 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0) x 335,885 ops/sec ±1.22% (92 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0) x 358,881 ops/sec ±0.89% (96 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true}) x 321,664 ops/sec ±0.85% (94 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true}) x 341,290 ops/sec ±1.00% (96 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true}) x 367,722 ops/sec ±0.95% (96 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true}) x 294,952 ops/sec ±0.56% (93 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true}) x 311,517 ops/sec ±0.36% (94 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true}) x 320,496 ops/sec ±1.18% (91 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 274,429 ops/sec ±0.58% (94 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 282,862 ops/sec ±0.72% (91 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 304,038 ops/sec ±0.44% (94 runs sampled)
```

I also added parseOptions before any method that passes `options` down to range or semver to reduce the number of comparisons by looking into `isParsedConfigSymbol`.

With this new way of parsing the options, the memory allocation is 0 for any parsed options.

The only downside is the maintenance, the number of variables to represent the options can increase depending on the number of options if we keep just 3 options, so is not a problem, if we add more, is more comparisons.

### Alternative Version

I also made a version using bit masks, which you can see in this commit: https://github.com/h4ad-forks/node-semver/commit/6c7a68ad9849cadbc624ef5208662ae1ea01d846

In that version, the performance benefits are the same, the maintenance is lower because is very easy to handle bit flags but is takes more memory and breaks one test of parse-options (about the number).

Also, the options now are just a getter and I introduce a new variable called `flagOptions`, which holds all the flags.

## Conclusion

In terms of lower changes and lower memory usage, the best option is `Object.freeze`.

If you want to take more aggressive changes with lower maintenance in the future and allocating some memory in the process, the option with bit masks is the best.

## References

Related to #528

<details>

```js
const satisfies = require('./functions/satisfies');

const versions = ['1.0.3||^2.0.0', '2.2.2||~3.0.0', '2.3.0||<4.0.0'];
const versionToCompare = '1.0.6';
const option1 = { includePrelease: true };
const option2 = { includePrelease: true, loose: true };
const option3 = { includePrelease: true, loose: true, rtl: true };

for (const version of versions) {
  suite.add(`satisfies(${versionToCompare}, ${version})`, function () {
    satisfies(versionToCompare, version);
  });
}

for (const version of versions) {
  suite.add(`satisfies(${versionToCompare}, ${version}, ${JSON.stringify(option1)})`, function () {
    satisfies(versionToCompare, version, option1);
  });
}

for (const version of versions) {
  suite.add(`satisfies(${versionToCompare}, ${version}, ${JSON.stringify(option2)})`, function () {
    satisfies(versionToCompare, version, option2);
  });
}

for (const version of versions) {
  suite.add(`satisfies(${versionToCompare}, ${version}, ${JSON.stringify(option3)})`, function () {
    satisfies(versionToCompare, version, option3);
  });
}

suite
  })
  .run({ async: false });
```

</details/>

### DIFF
--- a/bin/semver.js
+++ b/bin/semver.js
@@ -24,6 +24,7 @@ let rtl = false
 let identifier
 
 const semver = require('../')
+const parseOptions = require('../internal/parse-options')
 
 let reverse = false
 
@@ -88,7 +89,7 @@ const main = () => {
     }
   }
 
-  options = { loose: loose, includePrerelease: includePrerelease, rtl: rtl }
+  options = parseOptions({ loose, includePrerelease, rtl })
 
   versions = versions.map((v) => {
     return coerce ? (semver.coerce(v, options) || { version: v }).version : v
--- a/internal/parse-options.js
+++ b/internal/parse-options.js
@@ -1,11 +1,51 @@
-// parse out just the options we care about so we always get a consistent
-// obj with keys in a consistent order.
-const opts = ['includePrerelease', 'loose', 'rtl']
-const parseOptions = options =>
-  !options ? {}
-  : typeof options !== 'object' ? { loose: true }
-  : opts.filter(k => options[k]).reduce((o, k) => {
-    o[k] = true
-    return o
-  }, {})
+// parse out just the options we care about
+const isParsedConfigSymbol = Symbol('isParsedConfig')
+const var1 = Object.freeze({ includePrerelease: true, loose: true, rtl: true, [isParsedConfigSymbol]: true })
+const var2 = Object.freeze({ includePrerelease: true, loose: true, [isParsedConfigSymbol]: true })
+const var3 = Object.freeze({ includePrerelease: true, rtl: true, [isParsedConfigSymbol]: true })
+const var4 = Object.freeze({ includePrerelease: true, [isParsedConfigSymbol]: true })
+const var5 = Object.freeze({ loose: true, rtl: true, [isParsedConfigSymbol]: true })
+const var6 = Object.freeze({ loose: true, [isParsedConfigSymbol]: true })
+const var7 = Object.freeze({ rtl: true, [isParsedConfigSymbol]: true })
+const emptyOpts = Object.freeze({ [isParsedConfigSymbol]: true })
+
+const parseOptions = options => {
+  if (!options) {
+    return emptyOpts
+  }
+
+  if (typeof options !== 'object') {
+    return var6
+  }
+
+  if (options[isParsedConfigSymbol]) {
+    return options
+  }
+
+  if (options.includePrerelease) {
+    if (options.loose && options.rtl) {
+      return var1
+    }
+
+    if (options.loose) {
+      return var2
+    }
+
+    if (options.rtl) {
+      return var3
+    }
+
+    return var4
+  } else if (options.loose) {
+    if (options.rtl) {
+      return var5
+    }
+
+    return var6
+  } else if (options.rtl) {
+    return var7
+  } else {
+    return emptyOpts
+  }
+}
 module.exports = parseOptions
--- a/internal/parse-options.js
+++ b/internal/parse-options.js
@@ -1,51 +1,15 @@
 // parse out just the options we care about
-const isParsedConfigSymbol = Symbol('isParsedConfig')
-const var1 = Object.freeze({ includePrerelease: true, loose: true, rtl: true, [isParsedConfigSymbol]: true })
-const var2 = Object.freeze({ includePrerelease: true, loose: true, [isParsedConfigSymbol]: true })
-const var3 = Object.freeze({ includePrerelease: true, rtl: true, [isParsedConfigSymbol]: true })
-const var4 = Object.freeze({ includePrerelease: true, [isParsedConfigSymbol]: true })
-const var5 = Object.freeze({ loose: true, rtl: true, [isParsedConfigSymbol]: true })
-const var6 = Object.freeze({ loose: true, [isParsedConfigSymbol]: true })
-const var7 = Object.freeze({ rtl: true, [isParsedConfigSymbol]: true })
-const emptyOpts = Object.freeze({ [isParsedConfigSymbol]: true })
-
+const looseOption = Object.freeze({ loose: true })
+const emptyOpts = Object.freeze({ })
 const parseOptions = options => {
   if (!options) {
     return emptyOpts
   }
 
   if (typeof options !== 'object') {
-    return var6
+    return looseOption
   }
 
-  if (options[isParsedConfigSymbol]) {
-    return options
-  }
-
-  if (options.includePrerelease) {
-    if (options.loose && options.rtl) {
-      return var1
-    }
-
-    if (options.loose) {
-      return var2
-    }
-
-    if (options.rtl) {
-      return var3
-    }
-
-    return var4
-  } else if (options.loose) {
-    if (options.rtl) {
-      return var5
-    }
-
-    return var6
-  } else if (options.rtl) {
-    return var7
-  } else {
-    return emptyOpts
-  }
+  return options
 }
 module.exports = parseOptions
--- a/classes/comparator.js
+++ b/classes/comparator.js
@@ -78,13 +78,6 @@ class Comparator {
       throw new TypeError('a Comparator is required')
     }
 
-    if (!options || typeof options !== 'object') {
-      options = {
-        loose: !!options,
-        includePrerelease: false,
-      }
-    }
-
     if (this.operator === '') {
       if (this.value === '') {
         return true
--- a/functions/parse.js
+++ b/functions/parse.js
@@ -4,8 +4,6 @@ const SemVer = require('../classes/semver')
 
 const parseOptions = require('../internal/parse-options')
 const parse = (version, options) => {
-  options = parseOptions(options)
-
   if (version instanceof SemVer) {
     return version
   }
@@ -18,6 +16,7 @@ const parse = (version, options) => {
     return null
   }
 
+  options = parseOptions(options)
   const r = options.loose ? re[t.LOOSE] : re[t.FULL]
   if (!r.test(version)) {
     return null
--- a/test/internal/parse-options.js
+++ b/test/internal/parse-options.js
@@ -27,5 +27,20 @@ t.test('objects only include truthy flags we know about, set to true', t => {
     rtl: true,
     includePrerelease: true,
   })
+  t.strictSame(parseOptions({ loose: true }), { loose: true })
+  t.strictSame(parseOptions({ rtl: true }), { rtl: true })
+  t.strictSame(parseOptions({ includePrerelease: true }), { includePrerelease: true })
+  t.strictSame(parseOptions({ loose: true, rtl: true }), { loose: true, rtl: true })
+  t.strictSame(parseOptions({ loose: true, includePrerelease: true }), { loose: true, includePrerelease: true })
+  t.strictSame(parseOptions({ rtl: true, includePrerelease: true }), { rtl: true, includePrerelease: true })
+  t.end()
+})
+
+t.test('should skip validation when options is already parsed', t => {
+  const options = { loose: true, rtl: true }
+  const parsedOptions = parseOptions(options)
+
+  t.equal(parseOptions(parsedOptions) === parsedOptions, true)
+  t.not(parseOptions(options) === parsedOptions, false)
   t.end()
 })
--- a/test/internal/parse-options.js
+++ b/test/internal/parse-options.js
@@ -18,29 +18,26 @@ t.test('truthy non-objects always loose mode, for backwards comp', t => {
   t.end()
 })
 
-t.test('objects only include truthy flags we know about, set to true', t => {
-  t.strictSame(parseOptions(/asdf/), {})
-  t.strictSame(parseOptions(new Error('hello')), {})
-  t.strictSame(parseOptions({ loose: true, a: 1, rtl: false }), { loose: true })
+t.test('any object passed is returned', t => {
+  t.strictSame(parseOptions(/asdf/), /asdf/)
+  t.strictSame(parseOptions(new Error('hello')), new Error('hello'))
+  t.strictSame(parseOptions({ loose: true, a: 1, rtl: false }), { loose: true, a: 1, rtl: false })
   t.strictSame(parseOptions({ loose: 1, rtl: 2, includePrerelease: 10 }), {
-    loose: true,
-    rtl: true,
-    includePrerelease: true,
+    loose: 1,
+    rtl: 2,
+    includePrerelease: 10,
   })
   t.strictSame(parseOptions({ loose: true }), { loose: true })
   t.strictSame(parseOptions({ rtl: true }), { rtl: true })
   t.strictSame(parseOptions({ includePrerelease: true }), { includePrerelease: true })
   t.strictSame(parseOptions({ loose: true, rtl: true }), { loose: true, rtl: true })
-  t.strictSame(parseOptions({ loose: true, includePrerelease: true }), { loose: true, includePrerelease: true })
-  t.strictSame(parseOptions({ rtl: true, includePrerelease: true }), { rtl: true, includePrerelease: true })
-  t.end()
-})
-
-t.test('should skip validation when options is already parsed', t => {
-  const options = { loose: true, rtl: true }
-  const parsedOptions = parseOptions(options)
-
-  t.equal(parseOptions(parsedOptions) === parsedOptions, true)
-  t.not(parseOptions(options) === parsedOptions, false)
+  t.strictSame(parseOptions({ loose: true, includePrerelease: true }), {
+    loose: true,
+    includePrerelease: true,
+  })
+  t.strictSame(parseOptions({ rtl: true, includePrerelease: true }), {
+    rtl: true,
+    includePrerelease: true,
+  })
   t.end()
 })


bin/semver.js
```diff
@@ -88,7 +89,7 @@ const main = () => {
     }
   }
 
-  options = { loose: loose, includePrerelease: includePrerelease, rtl: rtl }
+  options = parseOptions({ loose, includePrerelease, rtl })
```
ljharb, line 92: why create a new object here at all if the goal is to avoid new object creation?

internal/parse-options.js
```diff
@@ -1,11 +1,46 @@
...
-  }, {})
+// parse out just the options we care about
+const isParsedConfigSymbol = Symbol('isParsedConfig');
+const var1 = Object.freeze({ includePrerelease: true, loose: true, rtl: true, [isParsedConfigSymbol]: true })
+const var2 = Object.freeze({ includePrerelease: true, loose: true, [isParsedConfigSymbol]: true })
+const var3 = Object.freeze({ includePrerelease: true, rtl: true, [isParsedConfigSymbol]: true })
+const var4 = Object.freeze({ includePrerelease: true, [isParsedConfigSymbol]: true })
+const var5 = Object.freeze({ loose: true, rtl: true, [isParsedConfigSymbol]: true })
+const var6 = Object.freeze({ loose: true, [isParsedConfigSymbol]: true })
+const var7 = Object.freeze({ rtl: true, [isParsedConfigSymbol]: true })
```
ljharb, line 9: might be nice to have better variable names here :-)

ranges/max-satisfying.js
```diff
@@ -1,10 +1,12 @@
 const SemVer = require('../classes/semver')
 const Range = require('../classes/range')
+const parseOptions = require('../internal/parse-options')
 
 const maxSatisfying = (versions, range, options) => {
   let max = null
   let maxSV = null
   let rangeObj = null
+  options = parseOptions(options)
```
ljharb, line 9: reassigning arguments is a deopt in older v8 engines, and potentially in other non-v8 engines as well; can we make a new variable instead?

bin/semver.js
H4ad: This method executes only once, I don't think is a problem to keep it this way, also, I put parseOptions here to reduce the number of calls for `parseOptions` using the original object (which adds more comparisons).

ranges/max-satisfying.js
H4ad: This deopt is true only for V8 engines, and we need to manipulate `arguments` in the method to cause the deopt, [ref](https://github.com/NathanaelA/v8-Natives/blob/master/opti-killers.md#3-managing-arguments).

H4ad:
Just to bring more attention, @ljharb points out a very interesting thing that could affect this PR at https://github.com/npm/node-semver/pull/536#pullrequestreview-1374076868.

 If the assumption is true, we can clean all the `varN` variables and just return the original object options.
This will only affect the tests, so we will need to fix them.

jakebailey:
Here's my DT stress test using pnpm. Before:

```
Done in 2m 14.5s
total time:  134.82s
user time:   162.81s
system time: 31.61s
CPU percent: 144%
max memory:  1752 MB
```

This PR (`"semver": "github:h4ad-forks/node-semver#fix/faster-parse-options"`):

```
Done in 1m 42.2s
total time:  102.43s
user time:   115.96s
system time: 30.80s
CPU percent: 143%
max memory:  1774 MB
```

Which according to https://how-much-faster.glitch.me/ is:

- 31.62% faster
- 1.32x faster
- 24.02% less time

jakebailey:
I forgot to say the obvious: well done!

Feel free to ping me if you want this rerun and I haven't already done so.

ranges/intersects.js
```diff
@@ -1,5 +1,7 @@
 const Range = require('../classes/range')
+const parseOptions = require('../internal/parse-options')
```
wraithgar, line 2: Do we really need to add this to the ranges modules?  With the exception of subset, none of the code there evaluates options, it simply passes it through.

I suppose if parseOptions is optimized enough that re-runs of it are cheaper than initial ones, we could see a net benefit by doing it up front like this.  It's at least worth considering what benefit this gives.

ranges/subset.js
```diff
@@ -3,6 +3,7 @@ const Comparator = require('../classes/comparator.js')
 const { ANY } = Comparator
 const satisfies = require('../functions/satisfies.js')
 const compare = require('../functions/compare.js')
+const parseOptions = require('../internal/parse-options.js')
```
wraithgar, line 6:
```sh
$ grep -r parse-options .|grep require|grep parse-options.js
./ranges/simplify.js:const parseOptions = require('../internal/parse-options.js')
./ranges/subset.js:const parseOptions = require('../internal/parse-options.js')
./test/internal/parse-options.js:const parseOptions = require('../../internal/parse-options.js')
```

Everywhere else requires `internal/parse-options` without `.js`

internal/parse-options.js
wraithgar: Agreed.  Starting from the perspective of a general roadmap of `make it, make it right, make it fast` this breaks `make it right` in that it is much much harder for a developer to follow.  I think in the balance between speed and developer experience this goes just a shade too far.

wraithgar:
Stepping back and broadly asking the question "what is `parseOptions` for"? I think we have the following:

1. standardize object for memoization
   - not needed thanks to #536)
2. coercion of `true|false` into `{ loose: true:false }`
   - required, semver major change to remove this
3. coercion of empty options into `{}`
   - not absolutely required, but definitely a best practice. Relying on the fact that `true` doesn't have a `loose` attribute is quite the antipattern
4. filtering out of unknown options
   - we already don't do anything if extra options are passed.
   - potentially a memory savings though?  Is the tradeoff worth it for the extra cpu time?

If all we did were `2` and `3` I think we would be ok?

ranges/intersects.js
H4ad: With the assumption of faster `parseOptions` indeed this call might be redundant, should I remove all those calls that `parseOptions` and just forward to Range/Semver?

ranges/subset.js
H4ad: I just keep the same pattern as other imports, I think we could leave it as it is and have another PR to standardize.

Because doesn't make too much sense to keep just parse-options without `.js` but is out of the scope to remove all imports from `.js`.

ranges/intersects.js
wraithgar: Yeah I think for now let's only call it if we actually interact with options directly.  We can revisit this decision in the future if `parseOptions` ever becomes a validator.

ranges/subset.js
ljharb: I'll be happy to make a separate PR removing the extension from all the imports :-)

jakebailey:
Personally, if I were to rewrite this library from scratch (something I'm trying to do to explore perf improvements, though I don't know if I will actually publish it), I wouldn't be storing these options in `SemVer` at all.

Looking at `SemVer`, only two options are used, `loose` and `includePrerelease`.

`loose` means that the parser can accept things like leading zeros. But once it's parsed, the loose-ness isn't important anymore. If it's being compared with another version, that's a different version, and the loose-ness of a differnet version is not important.

`includePrerelease`, on the other hand, isn't used at all. That option really only makes sense for `Range`. And even then, I still think that it's not a _property_ of a `Range`; it's a property of `satisfies` and other helpers and how they decide to _interpret_ a `Range`.

So to me, it feels like an equivalently good change to this PR would be to just check `loose` once, store it as a prop for people to know if a `SemVer` came from a loose parse, but otherwise not try and do anything.

wraithgar:
`rtl` is used in `coerce` and only there, and that `coerce` is not used by anything else in semver.

H4ad:
Well, I think we agree to keep the object as-is and don't do any modification, and just validate the cases of `undefined` and `boolean`.

In this case, I will create a new commit instead of modifying the first one with the test modified and the new behavior.

wraithgar:
> Also, this can be considered as breaking change?

It depends on what the changes needed are.  Just cause our tests have to change doesn't necessarily mean it's a breaking change.  If folks are going to get different results after this change by passing the exact same things as before, then it is a breaking change.

H4ad:
@wraithgar I thought it might be a breaking change if someone got the parsed object and access something in the options, but I don't think that will be the case, so maybe it's not a breaking change.

bin/semver.js
anonrig: I agree with @ljharb. If we are doing micro-optimizations, creating an extra object has a cost too. It would be faster to pass it as different arguments.

anonrig: I wonder if the following approach would be faster:

```
parseAndMutateOptions(options, loose, includePrerelease, rtl)
```

And instead of returning an object, and later copying it to a different variable, the function can directly mutate the `options` variable avoiding the extra copy.

H4ad: With the new behavior of returning the object in parseOptions, this function doesn't allocate extra memory and just returns the same object.

> parseAndMutateOptions(options, loose, includePrerelease, rtl)

From what I remember, this could be faster but also can suffer from deopt if some of those arguments are passed with different types, like `undefined`, `true`, etc...

> And instead of returning an object, and later copying it to a different variable, the function can directly mutate the options variable avoiding the extra copy.

Usually, this is a good approach but I don't think that is a good behavior to introduce without marking it as breaking change, but I will keep it in mind in case we are forced to return booleans.

jakebailey: Mutating the options seems like a bad idea; it seems quite possible that those options to have come from something that the caller thought would remain unmodified.

wraithgar: In light of https://github.com/npm/node-semver/pull/535#issuecomment-1499331981 I think we can leave bin/semver.js alone. It's fine as-is and isn't exactly a hot path.

wraithgar:
I think once the linting errors are fixed we are in a good place here.

@H4ad how do you feel about this PR in contrast to your other idea of using bitmasks?  Do you still think that is worth exploring?

wraithgar:
## `main`

```
satisfies(1.0.6, 1.0.3||^2.0.0) x 365,321 ops/sec ±0.83% (90 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0) x 380,982 ops/sec ±0.56% (93 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0) x 394,928 ops/sec ±0.92% (95 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true}) x 344,723 ops/sec ±0.85% (93 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true}) x 349,690 ops/sec ±1.50% (86 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true}) x 358,409 ops/sec ±2.81% (89 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true}) x 289,201 ops/sec ±1.26% (88 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true}) x 299,104 ops/sec ±1.22% (89 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true}) x 307,864 ops/sec ±1.27% (87 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 241,164 ops/sec ±1.90% (87 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 249,600 ops/sec ±1.52% (87 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 259,557 ops/sec ±1.50% (88 runs sampled)
```

## this PR as of [commit 585125](https://github.com/npm/node-semver/pull/535/commits/585125b6a4f35f345ae8d77ea9ca7cfe4ebbf999)

``` 
satisfies(1.0.6, 1.0.3||^2.0.0) x 397,658 ops/sec ±0.71% (85 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0) x 429,146 ops/sec ±0.69% (95 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0) x 446,333 ops/sec ±0.79% (93 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true}) x 344,022 ops/sec ±1.09% (92 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true}) x 354,640 ops/sec ±1.01% (89 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true}) x 364,129 ops/sec ±1.06% (92 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true}) x 314,245 ops/sec ±1.30% (86 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true}) x 321,380 ops/sec ±1.46% (90 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true}) x 338,411 ops/sec ±1.45% (89 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 284,783 ops/sec ±1.62% (85 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 274,453 ops/sec ±2.18% (83 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 306,652 ops/sec ±1.68% (84 runs sampled)
```

H4ad:
@wraithgar About bitmasks, I think it could be good for big refactorings of the library, for this PR, I don't think it's worth having that level of refactoring for the entire lib just to use bitmasks, the perf is almost the same as this current PR.

wraithgar:
Like #536 I'm giving this a 👍 and letting it sit through the weekend to let folks catch up.

This PR should definitely *not* land before #536, it is predicated off the fact that the parsed options are no longer being used to memoize a cache.

nlf (approved):

PR #538, opened 2023-04-06
fix: intersects with v0.0.0 and v0.0.0-0

wraithgar:
Also fix options not being passed through from semver.intersects

Closes https://github.com/npm/node-semver/issues/521

### DIFF
--- a/classes/comparator.js
+++ b/classes/comparator.js
@@ -90,6 +90,16 @@ class Comparator {
       return new Range(this.value, options).test(comp.semver)
     }
 
+    // Special cases where nothing can possibly be lower
+    if (options.includePrerelease &&
+      (this.value === '<0.0.0-0' || comp.value === '<0.0.0-0')) {
+      return false
+    }
+    if (!options.includePrerelease &&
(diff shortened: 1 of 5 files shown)


wraithgar:
This can't land before #535 since that removes the dead code that is causing coverage failures here.

classes/comparator.js
```diff
@@ -97,6 +97,16 @@ class Comparator {
...
+      return false
+    }
+    if (!options.includePrerelease &&
+      (this.value.startsWith('<0.0.0') || comp.value.startsWith('<0.0.0'))) {
+      return false
+    }
+
     const sameDirectionIncreasing =
       (this.operator === '>=' || this.operator === '>') &&
       (comp.operator === '>=' || comp.operator === '>')
```
H4ad, line 112: It is not worth adding in this PR also the early returns mentioned here: https://github.com/npm/node-semver/pull/528#issuecomment-1499241387?

wraithgar: Sure, once I've gotten any feedback on the implementation of the bugfix we can add that as a separate commit.

wraithgar: Much cleaner now.

PR #539, opened 2023-04-07
fix: faster diff

H4ad:
Based on the comment here: https://github.com/npm/node-semver/pull/533#issuecomment-1499798190

The performance before:

```
diff(0.0.1, 0.0.1-pre) x 310,595 ops/sec ±1.86% (94 runs sampled)
diff(0.0.1, 0.0.1-pre-2) x 317,054 ops/sec ±0.33% (92 runs sampled)
diff(1.1.0, 1.1.0-pre) x 304,229 ops/sec ±1.80% (86 runs sampled)
```

After:

```
diff(0.0.1, 0.0.1-pre) x 480,421 ops/sec ±0.86% (96 runs sampled)
diff(0.0.1, 0.0.1-pre-2) x 471,080 ops/sec ±0.99% (94 runs sampled)
diff(1.1.0, 1.1.0-pre) x 476,067 ops/sec ±0.98% (95 runs sampled)
```

<details>

```js
const diff = require('./functions/diff')

const cases = [
  ['0.0.1', '0.0.1-pre', 'patch'],
  ['0.0.1', '0.0.1-pre-2', 'patch'],
  ['1.1.0', '1.1.0-pre', 'minor'],
]

for (const [v1, v2] of cases) {
  suite.add(`diff(${v1}, ${v2})`, function () {
    diff(v1, v2)
  })
}

suite
  })
  .run({ async: false })
```

</details>

### DIFF
--- a/functions/diff.js
+++ b/functions/diff.js
@@ -2,11 +2,11 @@ const parse = require('./parse')
 const eq = require('./eq')
 
 const diff = (version1, version2) => {
-  if (eq(version1, version2)) {
+  const v1 = parse(version1)
+  const v2 = parse(version2)
+  if (eq(v1, v2)) {
     return null
   } else {
-    const v1 = parse(version1)
-    const v2 = parse(version2)
     const hasPre = v1.prerelease.length || v2.prerelease.length
     const prefix = hasPre ? 'pre' : ''
     const defaultResult = hasPre ? 'prerelease' : ''


wraithgar (approved):

wraithgar:
This one is very small and clearly an improvement, so I'm merging it now.

PR #546, opened 2023-04-11
Fix incorrect result from diff sometimes with prerelease versions

tjenkinson:
The following inputs previously gave the incorrect result based on my understanding:

- `0.0.2-1` => `0.0.3` was `prepatch` when it should be `patch`, as it's doing a patch operation that goes from the first to the second. Fixed with first commit.
- `0.0.0-1` => `0.0.0` was `prerelease` when it should be `major`. I think `0.0.0-1` => `0.0.0` should be consistent with `1.0.0-1` => `1.0.0`. Fixed with second commit. This makes it more consistent but not sure if it's actually what people would expect. See below.
-  `1.0.0-1` => `2.0.0-1` was `premajor` but should be `major` because the biggest change is the major version. Fixed with third commit.

I'm actually not convinced the behaviour for the second point is correct. For any version change where the stable part remains the same and the prerelease gets removed, `npm version patch` would give that result. So should the diff always be `patch` for these cases?

I.e `1.1.0-pre` => `1.1.0` is achievable with `npm version patch` and `npm version minor`. I personally would have expected the diff between those to be `patch`, just like it would be if it was `1.1.1-pre` => `1.1.1`


functions/diff.js
```diff
@@ -1,39 +1,53 @@
...
-        return 'minor'
-      }
-      if (v1.major) {
-        return 'major'
-      }
-    }
-    return defaultResult // may be undefined
   }
+
+  const v1Higher = gt(v1, v2)
```
wraithgar, line 12: This reinstantiates the semver objects.

```suggestion
  const v1Higher = v1.compare(v2) > 0
```

In fact if you rebase this against `main` you'll see the equality check was changed to a `.compare` call already, we can move that to a variable and reuse it.

wraithgar:
Yeah I think the `pre*` diff results are very counterintuitive.  They are trying to leverage the operations that `.inc` would do, but diff is not currently directional.  Your PR attempts to make the comparison directional which is the part that would need to be discussed.

functions/diff.js
```diff
@@ -1,39 +1,53 @@
...
+  const highVersion = v1Higher ? v1 : v2
+  const lowVersion = v1Higher ? v2 : v1
+  const highHasPre = !!highVersion.prerelease.length
+  const lowHasPre = !!lowVersion.prerelease.length
+  const prefix = highHasPre && !lowHasPre ? 'pre' : ''
```
wraithgar, line 17: This is where a lot of magic now lies and could definitely use a comment.

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
...
     ['1.1.0-pre-1', '1.1.0-pre-2', 'prerelease'],
     ['1.0.0', '1.0.0', null],
+    ['1.0.0-1', '1.0.0-1', null],
     ['0.0.2-1', '0.0.2', 'patch'],
+    ['0.0.2-1', '0.0.3', 'patch'],
+    ['0.0.2-1', '0.1.0', 'minor'],
+    ['0.0.2-1', '1.0.0', 'major'],
     ['0.1.0-1', '0.1.0', 'minor'],
     ['1.0.0-1', '1.0.0', 'major'],
-    ['0.0.0-1', '0.0.0', 'prerelease'],
```
wraithgar, line 25: Is this the only change in behavior that's currently covered in tests?

functions/diff.js
tjenkinson: Thanks done this in the merge in

functions/diff.js
tjenkinson: I’ve added one. Look ok?

wraithgar: See that's exactly why we need that comment.  In my mind "high" prioritizes before "low" but semver is considering going from low to high.  Great comment, thanks.

test/functions/diff.js
tjenkinson: The new ones I added in the first commit had different values before the fix as they were classed as pre* versions before

wraithgar:
Gonna kind of spam those test lines w/ comments sorry for the noise but this deserves a good paper trail

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
     ['1.1.0', '1.1.0-pre', 'minor'],
     ['1.1.0-pre-1', '1.1.0-pre-2', 'prerelease'],
     ['1.0.0', '1.0.0', null],
+    ['1.0.0-1', '1.0.0-1', null],
     ['0.0.2-1', '0.0.2', 'patch'],
+    ['0.0.2-1', '0.0.3', 'patch'],
```
wraithgar, line 24: The difference is actually two .inc` patches, but that's still a `patch` difference.

```
> semver.inc('0.0.2-1', 'patch')
'0.0.2'
> semver.inc('0.0.2', 'patch')
'0.0.3'
```

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
...
+    ['0.0.2-1', '0.1.0', 'minor'],
```
wraithgar, line 25:
```
> semver.inc('0.0.2-1', 'minor')
'0.1.0'
```

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
...
+    ['0.0.2-1', '1.0.0', 'major'],
```
wraithgar, line 26:
```
> semver.inc('0.0.2-1', 'major')
'1.0.0'
```

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
     ['1.1.0', '1.1.0-pre', 'minor'],
     ['1.1.0-pre-1', '1.1.0-pre-2', 'prerelease'],
     ['1.0.0', '1.0.0', null],
+    ['1.0.0-1', '1.0.0-1', null],
```
wraithgar, line 22:
```
> semver.eq('1.0.0-1', '1.0.0-1')
true
```

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
...
     ['0.0.2-1', '0.0.2', 'patch'],
+    ['0.0.2-1', '0.0.3', 'patch'],
+    ['0.0.2-1', '0.1.0', 'minor'],
+    ['0.0.2-1', '1.0.0', 'major'],
     ['0.1.0-1', '0.1.0', 'minor'],
     ['1.0.0-1', '1.0.0', 'major'],
-    ['0.0.0-1', '0.0.0', 'prerelease'],
+    ['1.0.1-1', '1.0.1', 'patch'],
```
wraithgar, line 29:
```
> semver.inc('1.0.1-1', 'patch')
'1.0.1'
```

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
...
+    ['0.0.0-1', '0.0.0', 'major'],
```
wraithgar, line 30:
```
> semver.inc('0.0.0-1', 'major')
'0.0.0'
```

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
...
+    ['1.0.0-1', '2.0.0', 'major'],
```
wraithgar, line 31: Again this is two major `.inc` operations but that's still a major diff

```
> semver.inc('1.0.0-1', 'major')
'1.0.0'
> semver.inc('1.0.0', 'major')
'2.0.0'
```

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
...
+    ['1.0.0-1', '2.0.0-1', 'major'],
+    ['1.0.0-1', '1.1.0-1', 'minor'],
+    ['1.0.0-1', '1.0.1-1', 'patch'],
```
wraithgar, line 34: This is the only part I'm not sure about

```
> semver.inc('1.0.0-1', 'premajor')
'2.0.0-0'
```

wraithgar:
> `1.0.0-1` => `2.0.0-1` was premajor but should be major because the biggest change is the major version. Fixed with third commit.

This still seems like premajor as per my comment on the test fixture:

```javascript
> semver.inc('1.0.0', 'premajor')
'2.0.0-0'
```

wraithgar:
When using `.inc` any directive that begins with `pre` leaves you in a prerelease version.  `pre` does seem to always indicate that the new version is a prerelease.

test/functions/diff.js
tjenkinson: these are `pre*` now

test/functions/diff.js
tjenkinson: these are `pre*` now

functions/diff.js
```diff
@@ -3,36 +3,52 @@ const parse = require('./parse')
...
+    return 'patch'
+  }
+
+  if (lowVersion.minor) {
+    // anything higher than a minor bump would result in the wrong version
+    return 'minor'
+  }
+
+  // bumping major/minor/patch all have same result
+  return 'major'
```
tjenkinson, line 51: Related to the last part of the PR description, this would make more sense to me if this was all replaced with `return 'patch'`

The way `npm version` behaves with major and minor bumps when the current version is a prerelease and parts are `0` is a bit confusing to me

Specifically I don't understand why there is a 0 special case [here](https://github.com/npm/node-semver/blob/27817677794f592b592bf6181a80a4824ff762b2/classes/semver.js#L231) and [here](https://github.com/npm/node-semver/blob/27817677794f592b592bf6181a80a4824ff762b2/classes/semver.js#L215-L219)

wraithgar: [special case for inc minor](https://github.com/npm/node-semver/blob/27817677794f592b592bf6181a80a4824ff762b2/classes/semver.js#L231)

The comments tell the story there:
>        // If this is a pre-minor version, bump up to the same minor version.
>        // Otherwise increment minor.

A preminor version is a version whose patch is `0` and has a prerelease identifier.  Bumping a minor from that is a matter of dropping the prerelease identifier.

[special case for inc major
](https://github.com/npm/node-semver/blob/27817677794f592b592bf6181a80a4824ff762b2/classes/semver.js#L215-L219)

Same as before, only we are looking at minor and patch.  A premajor version is a version whose minor and patch are `0` and has a prerelease identifier.  Bumping a major from that is a matter of dropping the prerelease identifier.

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
...
+    ['0.0.2-1', '0.0.3', 'patch'],
+    ['0.0.2-1', '0.1.0', 'minor'],
+    ['0.0.2-1', '1.0.0', 'major'],
     ['0.1.0-1', '0.1.0', 'minor'],
     ['1.0.0-1', '1.0.0', 'major'],
-    ['0.0.0-1', '0.0.0', 'prerelease'],
+    ['1.0.1-1', '1.0.1', 'patch'],
+    ['0.0.0-1', '0.0.0', 'major'],
+    ['1.0.0-1', '2.0.0', 'major'],
+    ['1.0.0-1', '2.0.0-1', 'premajor'],
```
wraithgar, line 32:
```
> semver.inc('1.0.0-1', 'premajor')
'2.0.0-0'
```

functions/diff.js
tjenkinson: I guess it's a bit strange that you can go from `1.0.1-0` to `1.1.0` with `npm version minor` but you can't go from `1.0.0-0` to `1.1.0` because the command works differently for that case.

I still don't really get why 0 is special 🤷. Seems a bit inconsistent

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
...
+    ['1.0.0-1', '1.1.0-1', 'preminor'],
```
wraithgar, line 33:
```
> semver.inc('1.0.0-1', 'preminor')
'1.1.0-0'
```

test/functions/diff.js
```diff
@@ -19,10 +19,19 @@ test('diff versions test', (t) => {
...
+    ['1.0.0-1', '1.0.1-1', 'prepatch'],
```
wraithgar, line 34:
```
> semver.inc('1.0.0-1', 'prepatch')
'1.0.1-0'
```

functions/diff.js
wraithgar: It does appear inconsistent but that is due to the nature of what `1.0.0-0` is. It's the prerelease for a semver major version.  Dropping that prerelease is a major, there is no smaller increment you can make outside of increasing the prerelase identifier.

tjenkinson: Right yes that makes sense. The thing that's confusing I think is that 0.1.0 to 1.0.0 is concretely major, but I'm not sure going from 1.0.0-0 to 1.0.0 could be classed as a major change, it's just going prerelease to stable. Maybe patch/minor/major doesn't really make sense as a thing when going prerelease to stable 🤷

I think the difference between 1.0.1-0 and 1.0.1 has the same weight

tjenkinson: It does make sense if the result from diff when we're going from prerelease to stable release is meant to represent the diff between the previous stable version and the new one 💡, but that's a step I wouldn't expect the diff to be doing internally

wraithgar (approved):
Gonna let @nlf review this before landing

wraithgar:
<details>
<summary>diff.js</summary>

```javascript
const semver = require('.')
for (const level of ['major', 'premajor', 'minor', 'preminor', 'patch', 'prepatch', 'prerelease']) {
  console.log(`### ${level}`)
  for (const base of ['1.0.0', '1.0.0-pre', '1.2.0', '1.2.0-pre', '1.2.3', '1.2.3-pre', '1.2.3-0']) {
    console.group(`- ${base}`)
    const inc = semver.inc(base, level)
    const diff = semver.diff(base, inc)
    console.log(`inc: ${inc}`)
    console.log(`diff: ${diff}`)
    console.groupEnd(`- ${base}`)
  }
}
```
</details>

### major
- 1.0.0
  inc: 2.0.0
  diff: major
- 1.0.0-pre
  inc: 1.0.0
  diff: major
- 1.2.0
  inc: 2.0.0
  diff: major
- 1.2.0-pre
  inc: 2.0.0
  diff: major
- 1.2.3
  inc: 2.0.0
  diff: major
- 1.2.3-pre
  inc: 2.0.0
  diff: major
- 1.2.3-0
  inc: 2.0.0
  diff: major
### premajor
- 1.0.0
  inc: 2.0.0-0
  diff: premajor
- 1.0.0-pre
  inc: 2.0.0-0
  diff: premajor
- 1.2.0
  inc: 2.0.0-0
  diff: premajor
- 1.2.0-pre
  inc: 2.0.0-0
  diff: premajor
- 1.2.3
  inc: 2.0.0-0
  diff: premajor
- 1.2.3-pre
  inc: 2.0.0-0
  diff: premajor
- 1.2.3-0
  inc: 2.0.0-0
  diff: premajor
### minor
- 1.0.0
  inc: 1.1.0
  diff: minor
- 1.0.0-pre
  inc: 1.0.0
  diff: major
- 1.2.0
  inc: 1.3.0
  diff: minor
- 1.2.0-pre
  inc: 1.2.0
  diff: minor
- 1.2.3
  inc: 1.3.0
  diff: minor
- 1.2.3-pre
  inc: 1.3.0
  diff: minor
- 1.2.3-0
  inc: 1.3.0
  diff: minor
### preminor
- 1.0.0
  inc: 1.1.0-0
  diff: preminor
- 1.0.0-pre
  inc: 1.1.0-0
  diff: preminor
- 1.2.0
  inc: 1.3.0-0
  diff: preminor
- 1.2.0-pre
  inc: 1.3.0-0
  diff: preminor
- 1.2.3
  inc: 1.3.0-0
  diff: preminor
- 1.2.3-pre
  inc: 1.3.0-0
  diff: preminor
- 1.2.3-0
  inc: 1.3.0-0
  diff: preminor
### patch
- 1.0.0
  inc: 1.0.1
  diff: patch
- 1.0.0-pre
  inc: 1.0.0
  diff: major
- 1.2.0
  inc: 1.2.1
  diff: patch
- 1.2.0-pre
  inc: 1.2.0
  diff: minor
- 1.2.3
  inc: 1.2.4
  diff: patch
- 1.2.3-pre
  inc: 1.2.3
  diff: patch
- 1.2.3-0
  inc: 1.2.3
  diff: patch
### prepatch
- 1.0.0
  inc: 1.0.1-0
  diff: prepatch
- 1.0.0-pre
  inc: 1.0.1-0
  diff: prepatch
- 1.2.0
  inc: 1.2.1-0
  diff: prepatch
- 1.2.0-pre
  inc: 1.2.1-0
  diff: prepatch
- 1.2.3
  inc: 1.2.4-0
  diff: prepatch
- 1.2.3-pre
  inc: 1.2.4-0
  diff: prepatch
- 1.2.3-0
  inc: 1.2.4-0
  diff: prepatch
### prerelease
- 1.0.0
  inc: 1.0.1-0
  diff: prepatch
- 1.0.0-pre
  inc: 1.0.0-pre.0
  diff: prerelease
- 1.2.0
  inc: 1.2.1-0
  diff: prepatch
- 1.2.0-pre
  inc: 1.2.0-pre.0
  diff: prerelease
- 1.2.3
  inc: 1.2.4-0
  diff: prepatch
- 1.2.3-pre
  inc: 1.2.3-pre.0
  diff: prerelease
- 1.2.3-0
  inc: 1.2.3-1
  diff: prerelease

wraithgar:
I'm happy to land this now.  Any problems that still exist are going to be with `inc` and not `diff`, now that we've shown that the results match (prepatch/prerelease subtleties notwithstanding).

nlf (approved):
this is so great. thank you @tjenkinson!

PR #548, opened 2023-04-12
feat: allow identifierBase to be false

### LINKED ISSUE #441: [BUG] Increase pre* with pre-release identifiers without build
# What / Why
the semver package should be able to execute a increment  premajor, preminor and prepatch for versioning schemes without a build number.
It for sure affects the standard maven versioning system (which is semver compliant).

# Examples
## Actual
0.0.1-SNAPSHOT premajor -> 1.0.0-SNAPSHOT.0
0.0.1-SNAPSHOT preminor -> 0.1.0-SNAPSHOT.0
0.0.1-SNAPSHOT prepatch -> 0.0.2-SNAPSHOT.0

## Expected
0.0.1-SNAPSHOT premajor -> 1.0.0-SNAPSHOT
0.0.1-SNAPSHOT preminor -> 0.1.0-SNAPSHOT
0.0.1-SNAPSHOT prepatch -> 0.0.2-SNAPSHOT

Currently this package always sets a dot-separated build. I t would be desirable if the behavior could be chosen, or if it would at least respect the current versioning scheme.

lsvalina:
## What
Added Disable Prerelease Identifier Base flag for creating prerelease identifiers without appending `.0`/`.1`
Improved `Prerelease Identifier Base` to use it when identifier is not set (1.0.0-{Identifier Base}) and adds description to cli help for `-n` flag
## Why
Version `1.0.0-something` is valid Semantic Version and its not necessary to have "build" number for identifiers that have only one/ first build, changes reflect that so if another prerelease is created with same identifier the "build" number will be append according Prerelease Identifier Base argument

## References
Fixes #441
Improves #532


wraithgar:
A "disable" boolean is often very confusing because it's turning something off when it's true.

We already have a flag to change this number, perhaps supporting `false` would suffice.

README.md
```diff
@@ -239,20 +239,40 @@ $ semver 1.2.4-beta.0 -i prerelease
 
 The method `.inc` takes an optional parameter 'identifierBase' string
 that will let you let your prerelease number as zero-based or one-based.
+If you provide false zero will be omitted.
```
wraithgar, line 242:
```suggestion
Set to `false` to omit the prerelease number altogether.
```

README.md
```diff
@@ -239,20 +239,40 @@ $ semver 1.2.4-beta.0 -i prerelease
...
 ```
 
+```bash
+$ semver 1.2.3 -i prerelease --preid beta -n false
+1.2.4-beta
+```
+
+
+if another version is created with same identifier then the prerelease
+number is used based on Prerelease Identifier Base
```
wraithgar, line 269: This is confusing, what does it mean?

wraithgar: I think perhaps for now the code example tells a better story than any comment can.

bin/semver.js
```diff
@@ -178,6 +178,13 @@ Options:
 --ltr
         Coerce version strings left to right (default)
 
+-n <base>
+        Prerelease Identifier Base
+        that will let you let your prerelease number as 
+        zero-based or one-based. If you provide false
+        zero will be omitted. If you do not specify 
+        this parameter, it will default to zero-based.
+
```
wraithgar, line 187:
```suggestion
        that will let you let your prerelease number as 
        zero-based or one-based. If you provide false
        zero will be omitted. If you do not specify 
        this parameter, it will default to zero-based.

```

classes/semver.js
```diff
@@ -246,9 +246,12 @@ class SemVer {
         break
       // This probably shouldn't be used publicly.
       // 1.0.0 'pre' would become 1.0.0-0 which is the wrong direction.
-      case 'pre':
+      case 'pre': {
+        identifierBase = identifierBase === 'false' ? false : identifierBase
```
wraithgar, line 250: The cli is the only time this is going to be a string `'false'` and it is the cli that should cast this to `0`, `1`, or `false`.

classes/semver.js
```diff
@@ -246,9 +246,13 @@ class SemVer {
         break
       // This probably shouldn't be used publicly.
       // 1.0.0 'pre' would become 1.0.0-0 which is the wrong direction.
-      case 'pre':
+      case 'pre': {
+        const base = identifierBase === false ||
+        identifierBase === 'false' ||
+        Number(identifierBase) ? 1 : 0
+
```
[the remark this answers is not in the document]
lsvalina: is it better like this, because it should handle string `'false'` to support cli

classes/semver.js
```diff
@@ -259,23 +262,26 @@ class SemVer {
...
-            this.prerelease.push(0)
+            this.prerelease.push(base)
           }
         }
         if (identifier) {
-          const base = Number(identifierBase) ? 1 : 0
           // 1.2.0-beta.1 bumps to 1.2.0-beta.2,
           // 1.2.0-beta.fooblz or 1.2.0-beta bumps to 1.2.0-beta.0
+          let prerelease = [identifier, base]
+          if (identifierBase === false || identifierBase === 'false') {
```
wraithgar, line 272: We don't need to support `'false'`, the cli should take care of this case.

README.md
lsvalina: I accidentally left outdated paragraph from initial commit, is this better

tap-snapshots/test/bin/semver.js.test.cjs
```diff
@@ -149,6 +154,11 @@ Object {
     --ltr
             Coerce version strings left to right (default)
     
+    -n <base>
+            Base number to be used for the prerelease identifier.
+            Can be either 0or 1, or false to omit the number altogether.
```
wraithgar, line 159:
```suggestion
            Can be either 0 or 1, or false to omit the number altogether.
```

bin/semver.js
```diff
@@ -178,6 +178,11 @@ Options:
 --ltr
         Coerce version strings left to right (default)
 
+-n <base>
+        Base number to be used for the prerelease identifier.
+        Can be either 0or 1, or false to omit the number altogether.
```
wraithgar, line 183:
```suggestion
        Can be either 0 or 1, or false to omit the number altogether.
```

test/fixtures/increments.js
```diff
@@ -101,4 +101,19 @@ module.exports = [
...
+  ['1.2.1', 'prerelease', '1.2.2-alpha', false, 'alpha', false],
+  ['1.2.2', 'prerelease', '1.2.3-alpha', false, 'alpha', false],
+  ['1.2.0', 'prepatch', '1.2.1-dev', false, 'dev', false],
+  ['1.2.0-1', 'prepatch', '1.2.1-dev', false, 'dev', false],
+  ['1.2.0', 'premajor', '2.0.0-dev', false, 'dev', false],
+  ['1.2.3-1', 'premajor', '2.0.0-dev', false, 'dev', false],
+  ['1.2.3-dev.bar', 'prerelease', '1.2.3-dev', false, 'dev', false],
+  ['1.2.0', 'preminor', '1.3.0-dev', false, 'dev', false],
+  ['1.2.3-1', 'preminor', '1.3.0-dev', false, 'dev', false],
+  ['1.2.3-dev', 'prerelease', '1.2.3-dev.1', false, 'dev', false],
```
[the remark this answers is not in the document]
nlf: my gut reaction is we should throw because they've asked for us to do something that doesn't make sense. going from no number to adding a number when we've been asked to not add numbers feels wrong

wraithgar:
Can you add the tests from https://github.com/npm/node-semver/pull/550 to this PR?

test/fixtures/increments.js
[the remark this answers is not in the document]
wraithgar: I agree, let's throw in this situation.  No more guessing user intent!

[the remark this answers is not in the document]
lsvalina: just to clarify do we want to throw also in:
```js
['1.2.0', 'prerelease', '1.2.1-1', false, '', false],
````
?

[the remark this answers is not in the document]
Comment (wraithgar) [reply test/fixtures/increments.js:115]: Oooh good point. We sure do!  "increase the prereleas with no identifier and no identifierBase" sure sounds impossible.

File: test/bin/semver.js [hunk 10]
```diff
@@ -28,8 +28,8 @@ const run = args => new Promise((resolve, reject) => {
 t.test('inc tests', t => Promise.all([
   ['-i', 'major', '1.0.0'],
   ['-i', 'major', '1.0.0', '1.0.1'],
-  ['-i', 'premajor', '1.0.0', '--preid=beta'],
```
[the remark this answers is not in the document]
lsvalina: no, i restore it, good catch

test/fixtures/increments.js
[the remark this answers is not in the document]
lsvalina: i updated it to throw on both cases

wraithgar:
Really useful new feature, thanks!

PR #551, opened 2023-04-13
Update map.js

SpriteyOnTheDraconic:
I replaced the arrow function with a named function for clarity and readability. I added a conditional statement to check if the testFile argument starts with the "test/" prefix before removing it. This avoids removing part of the file path if it doesn't contain the "test/" prefix. I defined the prefix variable to store the "test/" prefix for readability and maintainability. I moved the logic for removing the "test/" prefix to a separate function to make it reusable and testable. I exported the removeTestPrefix function at the end of the file to make it accessible to other modules.

### DIFF
--- a/map.js
+++ b/map.js
@@ -1,1 +1,8 @@
-module.exports = testFile => testFile.replace(/test\//, '')
+// Define a function that removes the "test/" prefix from a file path
+function removeTestPrefix(testFile) {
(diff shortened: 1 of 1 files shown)


wraithgar:
I think we're happy with where map.js is today, especially given that our direction in other repos has been to eventually remove that file.

PR #552, opened 2023-04-14
fix: throw on bad version with correct error message

wraithgar:
The parsing logic that is duplicated in SemVer's constructor was also removed.  See #541 for previous discussions about not optimizing for invalid SemVer.

With this fix:

```javascript
> semver.diff('foo', '1.2.3')
Uncaught TypeError: Invalid Version: foo
```

`v7.4.0`:

```javascript
> semver.diff('foo', '1.2.3')
Uncaught TypeError: Invalid Version: null
```

`503a4e52`:

```javscript
> semver.diff('foo', '1.2.3')
Uncaught TypeError: Cannot read properties of null (reading 'compare')
```

### DIFF
--- a/functions/parse.js
+++ b/functions/parse.js
@@ -1,22 +1,15 @@
-const { MAX_LENGTH } = require('../internal/constants')
 const SemVer = require('../classes/semver')
-const parse = (version, options) => {
+const parse = (version, options, throwErrors = false) => {
   if (version instanceof SemVer) {
     return version
   }
-
-  if (typeof version !== 'string') {
-    return null
-  }
-
-  if (version.length > MAX_LENGTH) {
-    return null
-  }
-
   try {
     return new SemVer(version, options)
   } catch (er) {
-    return null
+    if (!throwErrors) {
+      return null
+    }
+    throw er
   }
 }
 
--- a/classes/semver.js
+++ b/classes/semver.js
@@ -16,7 +16,7 @@ class SemVer {
         version = version.version
       }
     } else if (typeof version !== 'string') {
-      throw new TypeError(`Invalid Version: ${version}`)
+      throw new TypeError(`Invalid Version: ${require('util').inspect(version)}`)
     }
 
     if (version.length > MAX_LENGTH) {
--- a/functions/diff.js
+++ b/functions/diff.js
@@ -1,8 +1,8 @@
-const parse = require('./parse')
+const parse = require('./parse.js')
 
 const diff = (version1, version2) => {
-  const v1 = parse(version1)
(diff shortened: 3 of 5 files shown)


wraithgar:
This does reinstantiate, perhaps we need an unsafe parse (i.e. one that doesn't catch)

functions/parse.js
```diff
@@ -16,6 +16,9 @@ const parse = (version, options) => {
   try {
     return new SemVer(version, options)
   } catch (er) {
+    if (throwErrors) {
```
tjenkinson, line 19: I wonder if the `return null`'s above should also be errors that are thrown? Currently even with `throwErrors` as `true` there are still cases where `null` could come back

wraithgar: Very good point and in trying to fix it I am reminded of discussions from #541.  Perhaps the very tiny performance hit for bad semver shouldn't outweigh the benefits of having a single consistent place for semver to parse versions.

wraithgar:
Arrays and objects still throw an ambiguous message but is it worth bringing in something like `util.inspect`?

```javascript
> require('.').diff([], '1.3.2')
Uncaught TypeError: Invalid Version: 
```

```javascript
> require('.').diff({}, '1.3.2')
Uncaught TypeError: Invalid Version: [object Object]
```

vs

```
throw new TypeError(`Invalid Version: ${require('util').inspect(version)}`)`
```

```javascript
> require('.').diff([], '1.3.2')
Uncaught TypeError: Invalid Version: []
```

```javascript
> require('.').diff({}, '1.3.2')
Uncaught TypeError: Invalid Version: {}
```

functions/parse.js
tjenkinson: Hmm I was thinking keep the if statements but I’v move them into the try and have them throw an error instead of return null

If the performance overhead is the exception maybe it’s fine it’s just the error path?

tjenkinson: Oh I see the checks already exist so having them here was already redundant 

https://github.com/npm/node-semver/blob/503a4e52fe2b1c6ed1400d33149f7733c8361eed/classes/semver.js#L18-L26

nlf (approved):

UNIDY2002:
> I don't see any harm in making the error more readable with `inspect`

Oops... This does break something.

I am using this library in a react-native project, and since RN does not run in a Node.js environment, its bundler might not recognize the `"util"` module, causing [bundle failure](https://github.com/thu-info-community/thu-info-app/actions/runs/4725165235/jobs/8383254733).

I've just received a failure report, and haven't investigated into it.

Issue #554, opened 2023-04-18
[BUG] Webpack builds are failing as new release 7.5.0 requiers node.js util


janhommes:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

A webpack build is failing with the following error:

```
./node_modules/xxx/node_modules/semver/classes/semver.js:30:46-69 - Error: Module not found: Error: Can't resolve 'util' in '/runner/_work/xxx/node_modules/xxx/node_modules/semver/classes'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "util": require.resolve("util/") }'
	- install 'util'
If you don't want to include a polyfill, you can use an empty module like this:
	resolve.fallback: { "util": false }
```

It was working in version 7.4.0

### Expected Behavior

The webpack build should work without adding any node.js dependency as we want to use semver in the browser.

### Steps To Reproduce

1. Setup a webpack build
2. add semver
3. Run webpack
4. See error above

I assume it was introduced due to this change:
https://github.com/npm/node-semver/compare/v7.4.0...v7.5.0#diff-7917f8ef42ff71df5bde71c259d15feab8b147e3e6c2fde11b56cbf42cadea16R19

### Environment

- npm: 6.14.17
- Node: 14.9.3
- OS: Ubuntu 20.04
- platform: Linux

wraithgar:
This is obviously one of those "soft" areas of semver.  We have an `engines` declaration and test in all the versions laid out there.  Other environments are technically outside the support matrix.

Given the presence of tools like https://github.com/browserify/node-util specifically for things like this I wonder if fixing one's webpack setup is appropriate.

janhommes:
> This is obviously one of those "soft" areas of semver. We have an `engines` declaration and test in all the versions laid out there. Other environments are technically outside the support matrix.
> 
> Given the presence of tools like https://github.com/browserify/node-util specifically for things like this I wonder if fixing one's webpack setup is appropriate.

[...]

PR #559, opened 2023-05-05
fix: show type on invalid semver error

tjenkinson:
Based on [this comment](https://github.com/npm/node-semver/issues/554#issuecomment-1525740509) from @wraithgar 

> If we want to revert this, it needs to be done in a way that also sets up some sort of linting or other testing so that "does not use x internals from node" is part of the contract, which it wasn't before.

This configures some eslint plugins to prevent importing node builtins or dev dependencies.

It also removes the dependency on node `util` and updates the error message slightly.

## References
Fixes https://github.com/npm/node-semver/issues/554


.gitignore
```diff
@@ -33,3 +33,4 @@
 !/SECURITY.md
 !/tap-snapshots/
 !/test/
+!/rollup.config.js
```
tjenkinson, line 36: this change needs making in `@npmcli/template-oss`

wraithgar:
Is there really no eslint rule for this?  I find that amazing if so.

classes/semver.js
```diff
@@ -16,7 +16,7 @@ class SemVer {
         version = version.version
       }
     } else if (typeof version !== 'string') {
-      throw new TypeError(`Invalid Version: ${require('util').inspect(version)}`)
+      throw new TypeError(`Invalid Version: ${version}`)
```
wraithgar, line 19:
```
> x = []
[]
> throw new Error(`invalid version ${x}`)
Uncaught Error: invalid version 
> 
```

This tells the end user nothing valuable if the version is not a string.

wraithgar:
Is there something special about rollup or is it a coincidence that it also only supports the same subset of node internals that webpack does?

wraithgar:
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-nodejs-modules.md Is probably our rule. You'll need to override it in tests and map.js.

eslint rules can be extended by adding a `.eslintrc.local.js` file.

.eslintrc.local.js
```diff
@@ -0,0 +1,35 @@
+'use strict'
+const fs = require('fs')
+
+const pkgJson = JSON.parse(fs.readFileSync('./package.json', 'utf-8'))
+
+module.exports = {
+  plugins: ['import'],
+  rules: {
+    'import/no-extraneous-dependencies': [
```
tjenkinson, line 9: this may not be required. checking

wraithgar: It's already in the base ruleset https://github.com/npm/eslint-config/blob/9e87b58d9e049004c802239cab1fabbca67caf74/lib/index.js#L213

tjenkinson: Ah yep. I think this config is useful though? The default would allow importing from `devDependencies`

classes/semver.js
tjenkinson: How about `'Invalid Version. Must be a string. Got type "${typeof version}"'?

.eslintrc.local.js
wraithgar: The file is already imported as a plugin, so the extra dependency isn't needed.  I think a single new `rule` is all we need here with an `allow` of `./test` and `map.js`.

wraithgar: cf https://github.com/npm/cli/blob/3d5bbcc86c484ef0b6c04ee0236c1072d36915d6/.eslintrc.local.json#L3 or https://github.com/npm/cacache/blob/de12400af2ab452f1e75a15edaf7ea51b294667b/.eslintrc.local.js#L4

classes/semver.js
tjenkinson: Pushed something similar to this

.eslintrc.local.js
tjenkinson: Cool. Will remove the dependency. Feels a bit strange though not having it because without that the plugin is only there as a side effect of `@npmcli/template-oss` bringing it in

tjenkinson: Done that. If at some point the plugin is removed from `@npmcli/eslint-config` it looks like eslint will complain that the rule is missing

test/classes/semver.js
```diff
@@ -5,52 +5,60 @@ const comparisons = require('../fixtures/comparisons.js')
...
-    const s1 = new SemVer(v1, opt)
-    t.equal(s0.compare(s1), 1)
-    t.equal(s0.compare(v1), 1)
-    t.equal(s1.compare(s0), -1)
-    t.equal(s1.compare(v0), -1)
-    t.equal(s0.compare(v0), 0)
-    t.equal(s1.compare(v1), 0)
-    t.end()
-  }))
+  comparisons.forEach(([v0, v1, opt]) =>
```
ljharb, line 10: why the unrelated formatting changes?

tjenkinson: My ide automatically ran prettier I think, which fixed the line length for something eslint was complaining about, so I just kept the changes

I'll flip the rest of the lines back to what they were

tjenkinson: Yeh I did for the others but eslint couldn’t auto fix the indentation issue and I got lazy 😆

ljharb (commented):
An alternative would be to use the "browser" field to specify an alternative for `util.inspect` outside of node, which could be `typeof`, but then node usage would still get the nice util.inspect experience.

.eslintrc.local.js
wraithgar: I still think this config needs to be removed, it's part of our core linting already. We're only concerned about the nodejs modules rule.

ljharb:
¯\\\_(ツ)_/¯ ime the typeof the value isn't very useful when debugging a failure, but the inspect output is [very useful](https://www.npmjs.com/package/object-inspect).

.eslintrc.local.js
tjenkinson: But if someone imports something that’s not a dependency (or peer dep), that would cause errors for users?

I think the default config would not catch those

wraithgar: It will currently.  We have that rule enabled on all of the oss repos we manage through template-oss via the npm eslint config.


```sh
~/D/n/n/b/main (main|✔) $ npm run lint

> semver@7.5.0 lint
> eslint "**/*.js"


/Users/wraithgar/Development/npm/node-semver/branches/main/bin/test.js
  1:7   error  'nopt' is assigned a value but never used                                             no-unused-vars
  1:14  error  'nopt' should be listed in the project's dependencies. Run 'npm i -S nopt' to add it  import/no-extraneous-dependencies
```

tjenkinson: It's enabled, but the default config wouldn't catch an import from `devDependencies`

## Without config
image

## With config in the pr
image

wraithgar: Then this needs to be addressed in template-oss, that's not a rule that we'd want to only apply to a single one of our over 80 repos.

tjenkinson: Opened https://github.com/npm/template-oss/pull/316

tjenkinson: @wraithgar so given this doesn't follow the standard structure https://github.com/npm/eslint-config/pull/67 doesn't have any effect here, so is the custom config in this PR good to go?

wraithgar: Yeah we'll need a custom config but this one looks to be a bit different in approach than the one we landed on in eslint-config.  We should pattern it after the one there but instead of `lib` and `bin` it should be `bin`, `classes`, `functions`, `internal`, and `ranges` (and now we begin to see why we standardized on `lib`).

The way this PR currently does it is by completely bypassing the `no-extraneous-dependencies` rule for the test files, which is a regression.

tjenkinson: Yep. done!

wraithgar (approved):

classes/semver.js
```diff
@@ -16,7 +16,7 @@ class SemVer {
         version = version.version
       }
     } else if (typeof version !== 'string') {
-      throw new TypeError(`Invalid Version: ${require('util').inspect(version)}`)
+      throw new TypeError(`Invalid version. Must be a string. Got type "${typeof version}".`)
```
mbtools, line 19: "must be a string **or a SemVer**" would be more accurate

PR #566, opened 2023-06-15
fix: diff when detecting major change from prerelease

### LINKED ISSUE #561: [BUG] diff is inconsistent in comparing prerelease to later versions

### Current Behavior

This code:
```
    console.log(diff('2.0.0-alpha.1', '2.0.0'))
    console.log(diff('2.0.0-alpha.1', '2.1.1'))
```
Results in "major" and "minor".

### Expected Behavior

The code should result in the output "major", "major".

### Steps To Reproduce

Just run the code with semver v7.5.1. The bug seems to have appeared in 7.4.0.

### Environment

- npm: 8.18.0
- Node: v18.8.0
- OS: Windows 11
- platform: node

lukekarrys:
Fixes #561


functions/diff.js
```diff
@@ -13,6 +13,35 @@ const diff = (version1, version2) => {
...
+
+    // Otherwise it can be determined by checking the high version
+
+    if (highVersion.patch) {
+      // anything higher than a patch bump would result in the wrong version
+      return 'patch'
+    }
+
+    if (highVersion.minor) {
+      // anything higher than a patch bump would result in the wrong version
```
ljharb, line 38:
```suggestion
      // anything higher than a minor bump would result in the wrong version
```
?

wraithgar (approved):

PR #570, opened 2023-06-16
docs: add `pre` option for `inc`

### LINKED ISSUE #124: [DOCS] How do I get to 1.0.0-5 using increments?
I tried this:

```
> s = require('semver')
> v='0.0.0'
'0.0.0'
> v = s.inc(v, 'premajor')
'1.0.0-0'
> v = s.inc(v, 'premajor')
'2.0.0-0'
> v = s.inc(v, 'premajor')
'3.0.0-0'
> v = s.inc(v, 'premajor')
```

I would expect `1.0.0-0`, `1.0.0-1`, `1.0.0-2` and so on -- as I can have a number of pre-releases before I do `s.inc(v, 'major')`, right?

Is this just my misunderstanding or is it a bug? And if this behaviour is expected, how do I get to e.g. 1.0.0-5 using the `inc` function? Or should I use the **undocumented** `pre` option from https://github.com/npm/node-semver/blob/master/semver.js#L457 ?

Btw, the same applies to `inc` with `preminor` or `prepatch`.

mbtools:
Adds description of using `pre` with `inc` function with caveat that it might lead to unexpected results if version does not include a pre-release.

## References

Closes #124

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,8 +425,8 @@ strings that they parse.
     increments it.
   * If called from a prerelease version with `pre`, it increments the 
     revision number. Example: `inc('1.0.1-rc.1', 'pre') -> '1.0.1-rc.2'`
-    Be aware that `inc('1.0.0', 'pre') -> '1.0.0-0'` which is the wrong
-    direction.
+    Be aware that `inc('1.0.0', 'pre')` results in `1.0.0-0` which is 
+    the wrong direction.
 * `prerelease(v)`: Returns an array of prerelease components, or null
   if none exist. Example: `prerelease('1.2.3-alpha.1') -> ['alpha', 1]`
 * `major(v)`: Return the major version number.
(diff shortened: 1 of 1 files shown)


wraithgar:
https://github.com/npm/node-semver/blob/e7b78de06eb14a7fa2075cedf9f167040d8d31af/classes/semver.js#L247

I think we shouldn't call attention to this or suggest using it.  It's a very much internal tool.

mbtools:
Hmm. The pre option always has been and still is a public feature. Most certainly it's used as well since it provides a useful function that is otherwise not available. (I would also say Issac agreed with documenting it in the referenced issue but didn't get to it).

wraithgar:
Perhaps it needs a bigger disclaimer then, it is definitely a method that does not work how one would expect.

mbtools:
Maybe better to remove the option to call `inc` with `pre`. It doesn't work in the cli anyway. That's probably considered breaking, so something for the v8 list.

PR #571, opened 2023-06-22
fix: set max lengths in regex for numeric and build identifiers

lukekarrys:


wraithgar (approved):

test/integration/whitespace.js
```diff
@@ -7,33 +7,43 @@ const minVersion = require('../../ranges/min-version')
...
+  t.end()
+})
 
+test('range with 0', (t) => {
+  const r = `1.2.3 ${zeroLarge} <1.3.0`
+  t.throws(() => new Range(r).range)
+  t.equal(validRange(r), null)
+  t.throws(() => minVersion(r).version)
+  t.equal(minSatisfying(['1.2.3']), null)
+  t.equal(maxSatisfying(['1.2.3']), null)
```
mbtools, line 33: these should probably use `r` 😉

lukekarrys: Good catch! https://github.com/npm/node-semver/pull/574

ob1:
Hi :)
I think this introduced the bug described at https://github.com/npm/node-semver/issues/580 - would appreciate you taking a look 🙏

lukekarrys:
Thanks for the issue @ob1. I've got a fix for that now in #583.

PR #643, opened 2023-10-18
fix: correctly increase version for "pre" versions

fjcasternao:
We are using this plugin to bump versions every time we release using our CI. It happens for us that we use to create prereleases for QA to test and it can happen that we have several preminors before reaching the final one.

We realised, there were some issues with the version calculation on this specific scenario, so wanted to submit the fix.

Failing test reflecting the issue is the one on line 130, but the fix was also affecting some of the existing tests so I had to modify them also.

### DIFF
--- a/classes/semver.js
+++ b/classes/semver.js
@@ -31,6 +31,7 @@ class SemVer {
     // this isn't actually relevant for versions, but keep it so that we
     // don't run into trouble passing this.options around.
     this.includePrerelease = !!options.includePrerelease
+    this.isPrerelease = false
 
     const m = version.trim().match(options.loose ? re[t.LOOSE] : re[t.FULL])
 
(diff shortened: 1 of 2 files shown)


wraithgar:
The existing tests are correct, I'm not sure what this change is trying to fix.

`1.2.0-0` is the pre-release for `1.2.0` not `1.2.1`.  If I bump `1.2.0-0` by a patch I should go up to `1.2.0`.  There is a significant amount of prior discussion on this in these PRs:

- https://github.com/npm/node-semver/pull/546
- https://github.com/npm/node-semver/pull/548

fjcasternao:
Maybe I am wrong, but to me, `1.2.0-0` is a preminor bump from let’s say `1.1.9`

in https://github.com/npm/node-semver/pull/546 on your last comment you mention this `1.2.0-pre
inc: 1.2.1-0
diff: prepatch` that I guess is consistent with my approach here

PR #658, opened 2023-11-30
chore: add clean() test for build metadata

jethrodaniel:
Adds a test for `semver.clean` that demonstrates how the function strips build metadata.

### DIFF
--- a/test/functions/clean.js
+++ b/test/functions/clean.js
@@ -17,6 +17,7 @@ test('clean tests', (t) => {
     ['~1.2.3', null],
     ['<=1.2.3', null],
     ['1.2.x', null],
+    ['0.12.0-dev.1150+3c22cecee', '0.12.0-dev.1150'],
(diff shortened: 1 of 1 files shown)


jethrodaniel:
> Was there more you wanted to do here?

Nope, that's all.

Personally, I didn't realize that [semver](https://semver.org/#spec-item-10) considers versions that differ by only the build metadata to be equal in precedent... 

so when I noticed that `semver.clean` strips it in an unrelated project, the first place I came to look for clarification on the intended behavior was here, and then noticed that the tests alone couldn't confirm that for me.

I then went to reading semver spec more closely and everything made sense.

wraithgar:
> Nope, that's all.

You can mark this as ready then.

jethrodaniel:
> > Nope, that's all.
> 
> You can mark this as ready then.

Doh, forgot that it was in draft status. 🤦

PR #671, opened 2023-12-21
feat: preserve pre-release and build parts of a version on coerce (#592)

### LINKED ISSUE #357: [BUG] Filter version with prerelease tags
# What / Why
I'm trying to extract a version from string keeping prerelease tags but it's not considering it.

### Current Behavior
For example, running this code bellow, it should tell me that it's prerelease with a preview tag or just keep its integrity

```js
var semver = require("semver");
semver.coerce('installer-v1.66.18-preview.6', { includePrerelease: true, loose: true });
```
But it returns: 
```
build: []
includePrerelease: true
loose: true
major: 1
minor: 66
options: Object {includePrerelease: true, loose: true}
patch: 18
prerelease: []
raw: "1.66.18"
version: "1.66.18"
SemVer Prototype
```

### Expected Behavior
I do expect it to keep version like `1.66.18-preview.6` or at least tell me that it's a prerelease instead of getting rid of everything.

### LINKED ISSUE #592: `coerce` drops build and prerelease information
Hi there 👋 

I am currently using `semver.coerce` to coerce semver-like strings into a string which only contains a major, minor, and patch version.

The documentation doesn't mention anything about the other things which `semver.parse` may produce (build number, prerelease, etc.).
...is it safe for me to depend on this behaviour? Is there a better way to do this?

In addition, I need to coerce a lot of strings missing a patch version:

```
>require('semver').coerce('10.20-45')
SemVer {
  options: {},
  loose: false,
  raw: '10.20.0',
  major: 10,
  minor: 20,
  patch: 0,
  prerelease: [],
  build: [],
  version: '10.20.0'
}
```

The reason I am asking is because of somewhat surprising results. While I don't want the current behaviour of `semver` to change, I think if this is too surprising maybe it should. What do you think? 🤔 

```
Welcome to Node.js v16.20.0.
Type ".help" for more information.
> const semver = require('semver')
undefined
> const semverStringWithPrerelease = '10.20.30-40';
undefined
> semver.valid(semverStringWithPrerelease);
'10.20.30-40'
> semver.parse(semverStringWithPrerelease);
SemVer {
  options: { loose: false, includePrerelease: false },
  loose: false,
  raw: '10.20.30-40',
  major: 10,
  minor: 20,
  patch: 30,
  prerelease: [ 40 ],
  build: [],
  version: '10.20.30-40'
}
> semver.coerce(semverStringWithPrerelease);
SemVer {
  options: {},
  loose: false,
  raw: '10.20.30',
  major: 10,
  minor: 20,
  patch: 30,
  prerelease: [],
  build: [],
  version: '10.20.30'
}
> semver.compare(semverStringWithPrerelease, semverStringWithPrerelease)
0
> semver.compare(semver.coerce(semverStringWithPrerelease), semverStringWithPrerelease)
1
```

madtisa:
# What / Why

Introduces the new coerce option `includePrerelease`, if set, allowing to preserve pre-release and build parts of a version.

## References

Fixes #592
Fixes #357

### DIFF
--- a/functions/coerce.js
+++ b/functions/coerce.js
@@ -27,7 +27,7 @@ const coerce = (version, options) => {
     // With includePrerelease option set, '1.2.3.4-rc.1' wants to coerce '2.3.4-rc.1', not '1.2'
     //
     // Walk through the string checking with a /g regexp
-    // Manually set the index so, as to pick up overlapping matches.
+    // Manually set the index so as to pick up overlapping matches.
     // Stop when we get a match that ends at the string end, since no
     // coercible string can be more right-ward without the same terminus.
     const coerceRtlRegex = options.includePrerelease ? re[t.COERCERTLFULL] : re[t.COERCERTL]
--- a/functions/coerce.js
+++ b/functions/coerce.js
@@ -19,34 +19,42 @@ const coerce = (version, options) => {
 
   let match = null
   if (!options.rtl) {
-    match = version.match(re[t.COERCE])
+    match = version.match(options.full ? re[t.COERCEFULL] : re[t.COERCE])
   } else {
     // Find the right-most coercible string that does not share
     // a terminus with a more left-ward coercible string.
     // Eg, '1.2.3.4' wants to coerce '2.3.4', not '3.4' or '4'
+    // With full option set, '1.2.3.4-pre.1.2' wants to coerce '2.3.4-pre.1.2', not '1.2'
     //
     // Walk through the string checking with a /g regexp
-    // Manually set the index so as to pick up overlapping matches.
+    // Manually set the index so, as to pick up overlapping matches.
     // Stop when we get a match that ends at the string end, since no
     // coercible string can be more right-ward without the same terminus.
+    const coerceRtlRegex = options.full ? re[t.COERCERTLFULL] : re[t.COERCERTL]
     let next
-    while ((next = re[t.COERCERTL].exec(version)) &&
+    while ((next = coerceRtlRegex.exec(version)) &&
         (!match || match.index + match[0].length !== version.length)
     ) {
       if (!match ||
             next.index + next[0].length !== match.index + match[0].length) {
         match = next
       }
-      re[t.COERCERTL].lastIndex = next.index + next[1].length + next[2].length
+      coerceRtlRegex.lastIndex = next.index + next[1].length + next[2].length
     }
     // leave it in a clean state
-    re[t.COERCERTL].lastIndex = -1
+    coerceRtlRegex.lastIndex = -1
   }
 
   if (match === null) {
     return null
   }
 
-  return parse(`${match[2]}.${match[3] || '0'}.${match[4] || '0'}`, options)
+  const major = match[2]
+  const minor = match[3] || '0'
+  const patch = match[4] || '0'
+  const prerelease = options.full && match[5] ? `-${match[5]}` : ''
+  const build = options.full && match[6] ? `+${match[6]}` : ''
+
+  return parse(`${major}.${minor}.${patch}${prerelease}${build}`, options)
 }
 module.exports = coerce
--- a/functions/coerce.js
+++ b/functions/coerce.js
@@ -24,7 +24,7 @@ const coerce = (version, options) => {
     // Find the right-most coercible string that does not share
     // a terminus with a more left-ward coercible string.
     // Eg, '1.2.3.4' wants to coerce '2.3.4', not '3.4' or '4'
-    // With includePrerelease option set, '1.2.3.4-pre.1.2' wants to coerce '2.3.4-pre.1.2', not '1.2'
+    // With includePrerelease option set, '1.2.3.4-rc.1.2' wants to coerce '2.3.4-rc.1.2', not '1.2'
     //
     // Walk through the string checking with a /g regexp
     // Manually set the index so, as to pick up overlapping matches.
--- a/functions/coerce.js
+++ b/functions/coerce.js
@@ -24,7 +24,7 @@ const coerce = (version, options) => {
     // Find the right-most coercible string that does not share
     // a terminus with a more left-ward coercible string.
     // Eg, '1.2.3.4' wants to coerce '2.3.4', not '3.4' or '4'
-    // With includePrerelease option set, '1.2.3.4-rc.1' wants to coerce '2.3.4-rc.1', not '1.2'
+    // With includePrerelease option set, '1.2.3.4-rc' wants to coerce '2.3.4-rc', not '2.3.4'
     //
     // Walk through the string checking with a /g regexp
     // Manually set the index so as to pick up overlapping matches.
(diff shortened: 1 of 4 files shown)


wraithgar:
No it's the opposite. If true, `includePrerelease` means to include the prereelase portion when doing comparisons.  The default is `false`.

It seems to make sense to extend this boolean to also include the coercion function.  What we want to avoid is adding a new single-use boolean with an extremely generic name like "full".

functions/coerce.js
```diff
@@ -19,34 +19,42 @@ const coerce = (version, options) => {
...
+    match = version.match(options.includePrerelease ? re[t.COERCEFULL] : re[t.COERCE])
   } else {
     // Find the right-most coercible string that does not share
     // a terminus with a more left-ward coercible string.
     // Eg, '1.2.3.4' wants to coerce '2.3.4', not '3.4' or '4'
+    // With includePrerelease option set, '1.2.3.4-pre.1.2' wants to coerce '2.3.4-pre.1.2', not '1.2'
     //
     // Walk through the string checking with a /g regexp
-    // Manually set the index so as to pick up overlapping matches.
+    // Manually set the index so, as to pick up overlapping matches.
```
wraithgar, line 30: "[so as to](https://dictionary.cambridge.org/dictionary/english/so-as-to)" is a valid phrase, this comma is not warranted.

internal/re.js
```diff
@@ -159,7 +159,15 @@ createToken('COERCE', `${'(^|[^\\d])' +
               `(?:\\.(\\d{1,${MAX_SAFE_COMPONENT_LENGTH}}))?` +
               `(?:\\.(\\d{1,${MAX_SAFE_COMPONENT_LENGTH}}))?` +
               `(?:$|[^\\d])`)
+createToken('COERCEFULL', `${'(^|[^\\d])' +
```
wraithgar, line 162: If this is all duplicating `COERCE` and then appending new content, it should reuse `t.COERCE`.  You can look to things like `LOOSEPLAIN` as an example.

wraithgar:
This is looking really good, heading in the right direction.  Just some regex cleanup and an erroneous comment change.  The actual implementation in `functions/coerce.js` is really good, much easier to read with the parts broken up like that.

wraithgar:
> It's done, should I squash (rebase) everything into one commit at the end?

You don't have to, I will squash this when I merge it.

test/functions/coerce.js
```diff
@@ -110,13 +110,58 @@ test('coerce tests', (t) => {
...
   ]
-  coerceToValid.forEach(([input, expected, options]) => {
-    const msg = `coerce(${input}) should become ${expected}`
-    t.same((coerce(input, options) || {}).version, expected, msg)
+  coerceToValid.forEach(([input, expected, options, shouldParse]) => {
+    const coerceExpression = `coerce(${input}, ${JSON.stringify(options)})`
+    const coercedVersion = coerce(input, options) || {}
+    let expectedText = expected
+
+    if (shouldParse) {
```
wraithgar, line 147: This is surprising to me.  What is different about the new fixtures that is not present in the old ones that means we only do this to the new ones? What is this asserting specifically?  If this is important I would assume that all of the tests would want this assertion.

wraithgar: Would any of the existing fixtures fail this test? If not we can remove that flag altogether and run them all through both assertions.  Having that extra step if it's not needed is extra cognitive load for the folks who come after us trying to understand the tests.

test/functions/coerce.js
```diff
@@ -110,13 +110,53 @@ test('coerce tests', (t) => {
...
+    ['1.2.3.4/7-rc.5+rev.6', '7.0.0-rc.5+rev.6', { rtl: true, includePrerelease: true }],
   ]
   coerceToValid.forEach(([input, expected, options]) => {
-    const msg = `coerce(${input}) should become ${expected}`
-    t.same((coerce(input, options) || {}).version, expected, msg)
+    const coerceExpression = `coerce(${input}, ${JSON.stringify(options)})`
+    const coercedVersion = coerce(input, options) || {}
+    const expectedVersion = parse(expected)
+    t.equal(
+      expectedVersion.compare(coercedVersion) || expectedVersion.compareBuild(coercedVersion),
```
wraithgar, line 147: If we're expecting both of these to be `0` it's a little odd to be running them both via `||`.

wraithgar:
Thanks for your patience and responses getting this PR ready.  This module is one of the most highly downloaded packages on the entire registry and we try to be careful and purposeful when making changes.  We're definitely on the right track here and I think this is something that will be able to land soon.

wraithgar:
We have been the lack of including `includePrerelease` in a function as a bug, fixing it is not a breaking change.  We can leave this PR open for a bit to see if others agree or disagree.

There is still the one odd assertion in `test/functions/coerce.js` that asserts two different things joined by an `||` that seems off to me.

test/functions/coerce.js
wraithgar: You're asserting all of those are `t.equal` to `0` meaning they both always run.


```js

t.equal(y || x, 0) // y and x are always evaluated since if y were not 0 the test would throw
```

test/functions/coerce.js
```diff
@@ -110,13 +110,48 @@ test('coerce tests', (t) => {
...
   coerceToValid.forEach(([input, expected, options]) => {
-    const msg = `coerce(${input}) should become ${expected}`
-    t.same((coerce(input, options) || {}).version, expected, msg)
+    const coerceExpression = `coerce(${input}, ${JSON.stringify(options)})`
+    const coercedVersion = coerce(input, options) || {}
+    const expectedVersion = parse(expected)
+    t.ok(
+      expectedVersion.compare(coercedVersion) === 0 &&
+        expectedVersion.compareBuild(coercedVersion) === 0,
+      `${coerceExpression} should be equal to ${expectedVersion}`
```
wraithgar, line 149:
```suggestion
    t.equal(expectedVersion.compare(coercedVersion), 0,
      `${coerceExpression} should be equal to ${expectedVersion}`)
    t.equal(expectedVersion.compareBuild(coercedVersion), 0,
      `${coerceExpression} build should be equal to ${expectedVersion}`)
```

wraithgar (approved):

wraithgar:
Looks good, we can give this some time for other folks to weigh in on the semver breaking-ness of this.

ljharb:
I mean it’s definitely breaking since it’s not a new option - it’s just unclear whether anyone’s overarching workflow will break as a result.

Tobbe:
Came here because I just ran into this issue myself.

This is my current workaround. But it only works because I assume any version that specifies pre-release info is going to be valid.
```js
const v = semver.valid(minVersion) || semver.coerce(minVersion)
```

Looking forward to seeing this PR get merged!

wraithgar (approved):

wraithgar (approved):

PR #673, opened 2024-01-05
chore: node:test and node:assert

wraithgar:
This is a proof of concept intended to explore what it would look like
to use `node:test` and `node:assert` for testing.  For the most part the
assertions themselves translated nicely.

As part of the refactor, large fixtures were moved out of some test
files and into `./test/fixtures`.

The parts that are probably not ready are:

 - Code coverage.  This flag is still experimental in Node.js.  There is
   also no way to set a coverage level requirement, nor is there
   coverage mapping.  This is the only repo left that the npm cli team
   manages which still uses coverage mapping.  Whether or not we want to
   keep coverage mapping is a discussion that could take place.
 - Timeout.  There is a test that assumes the timeout is not Infinity.
   This is Node.js's default, and can only apparently be changed from
   tests that are invoked from `run()`.
 - Test runner.  This still uses `tap` for running the tests.  Running
   the tests purely from node would require having a centralized runner,
   or refactoring all the tests to invoke `run()` and set a timeout. It
   would also require solving the coverage problem.
 - Snapshots.  There is no built-in snapshot functionality in the
   Node.js test suite. There was only one test that really could
   arguably benefit from it: the help output.  The rest really were
   better suited for explicit checks.

Conclusion: `node:test` is not feature complete, and we can not cut over
to it fully.  This PR may be ok to land as-is, since `node:test` returns
tap-compatible results.  Whether or not we want to land it like this is
a discussion to be had.  At the very least this demonstrates the amount
of work needed to translate tests from `tap` to `node:test`


wraithgar:
Obviously this also doesn't work on older versions of Node.js because this module didn't exist yet in them.

ljharb:
I find this outcome unsurprising.

PR #681, opened 2024-02-07
feat: Add Semver.org regex as constant

earendil06:
This PR introduces a new constant `SEMVER_REGEX` and its value is the regex provided by Semver.org.


ljharb:
why? this package is a proper parser, a regex should only be used as a last resort.

wraithgar:
I don't think this is appropriate for this module to export.  This is not a definitive source of regexes to parse semver, it is the parser itself.

PR #689, opened 2024-03-15
fix: parsing of version prefix in ranges

mbtools:
Symptom:

The following "ranges" are currently considered valid: `> vvv1.2.3`, `> v==1.2.3`, `> v=vv==v1.x`

According to the readme, only a single `v` or `=` are permissible as a prefix (and stripped). However, the validation is not strict enough and allows nonsensical prefixes as well.

Note: There are also test cases using `v=` as a prefix although that's not documented. These shall remain valid.

Cause:

The version prefix is validated using regex `[v=\\s]*` which does not match the specs.

Fix:

Replace the prefix validation with regex `(?:v=|v|=)?` for strict mode, and `(?:v=|v|=)?\s*` in loose mode.

Tests:

The cases mentioned above are checked to be invalid in strict as well as loose mode.


wraithgar:
This is a great suggestion for the next semver major of semver.  We tend not to keep PRs open for pending breaking changes, but an issue is fine.  We can link to it from https://github.com/npm/node-semver/pull/689

Issue #501, opened 2022-10-31
[TRACKING] `semver@8`


darcyclarke:
### (WIP)

This is a tracking issue & WIP set of potential breaking changes, features & bugs which are expected to be fixed in the next major version of `semver`.

#### Breaking Changes
- [ ] [Remove preload](https://github.com/npm/node-semver/blob/940723d22bca824993627c45ac30dd3d2854b8cd/preload.js#L1)
- [ ] [Rewrite the cli](https://github.com/npm/node-semver/issues/499)
- [ ] Bump engines
- [ ] [Rearrange files for better import experience](https://github.com/npm/node-semver/issues/498)
- [ ] #418 
- [ ] #378 
- [ ] #691

#### Bugs
- [ ] #108 
- [ ] #381 
- [ ] #458 
- [x] #376 
- [ ] #354 
- [ ] #483 
- [ ] #345

budarin:
Consider storing versions internally not in a string representation

https://marvinh.dev/blog/speeding-up-javascript-ecosystem/
please look at the paragraph "The curious case of semver"

Mister-Hope:
Hi, is there any release table for v8? I believe this could be added.

Asking because I am trying to replace some outdated deps from repos I am maintaining. server v7 is requiring on a "very old" version of lru-cache.

7814222651:
Pay 💰 50000
Please please plz pls sir sir 😭🧠💔🙏🏻🙏🏻🤝🤝🫡📱
![Uploading BA16C54F-779C-425F-B499-5D20B5DFAEBE.jpeg…]()

7814222651:
![729A6456-DA74-4319-A327-AFBB1B4F896B](https://github.com/user-attachments/assets/eeeeff7f-8850-47b9-bdfb-b00835fdd398)

PR #695, opened 2024-04-07
chore!: bump lru-cache to 10.x & increase node engine

H4ad:
Trying to improve/solve the issue with dedupe on `cli`: https://github.com/npm/cli/issues/7350

I don't know if it's a good idea since `semver` is being used by a lot of projects and bumping the version of `node-lru-cache` requires the bump of `engines.node` to `>=16` , dropping a lot of node versions at once.

### DIFF
--- a/classes/range.js
+++ b/classes/range.js
@@ -199,7 +199,7 @@ class Range {
 module.exports = Range
 
 const LRU = require('lru-cache')
-const cache = new LRU({ max: 1000 })
+const cache = new LRU.LRUCache({ max: 1000 })
 
 const parseOptions = require('../internal/parse-options')
 const Comparator = require('./comparator')
--- a/package.json
+++ b/package.json
@@ -45,10 +45,10 @@
     ]
   },
   "engines": {
-    "node": ">=10"
+    "node": ">=16"
   },
   "dependencies": {
-    "lru-cache": "^6.0.0"
+    "lru-cache": "^10.2.0"
   },
   "author": "GitHub Inc.",
   "templateOSS": {


wraithgar:
This would need to be part of https://github.com/npm/node-semver/issues/501 since changing engines is a semver major.

domoritz:
Node 16 is [not supported anymore](https://nodejs.org/en/about/previous-releases) so maybe this doesn't need to be a breaking change?

wraithgar:
Changing the engines to drop support for any node version, supported or not, is a breaking change for a package.

domoritz:
Okay, a major version bump makes sense then.

mbtools:
This package doesn't use any fancy features of `lru-cache`. It's just caching up to 1000 parses. 

How about we get rid of the dependency here and replace it with a simple class based on Map?

```js
class LRUCache {
    constructor(max = 0) {
        if (!Number.isInteger(max) || max < 0)
            throw new TypeError('max must be a nonnegative integer');

        this.max = max;
        this.map = new Map();
    }

    get(key) {
        const value = this.map.get(key);
        if (value === undefined)
            return undefined;
        else {
            // Remove the key from the map and add it to the end
            this.map.delete(key);
            this.map.set(key, value);
            return value;
        }
    }

    has(key) {
        return this.map.has(key);
    }

    delete(key) {
        if (this.map.has(key)) {
            this.map.delete(key);
            return true;
        }
        else
            return false;
    }

    set(key, value) {
        const deleted = this.delete(key);

        if (!deleted && value !== undefined) {
            // If cache is full, delete the least recently used item
            if (this.map.size >= this.max) {
                const firstKey = this.map.keys().next().value;
                this.delete(firstKey);
            }

            this.map.set(key, value);
        }

        return this;
    }

    clear() {
        this.map.clear();
    }

    capacity() {
        return this.max;
    }

    size() {
        return this.map.size;
    }

    *entries() {
        for (const [key, value] of this.map)
            yield [key, value];
    }
}

export default LRUCache;
```

I'm happy to provide a PR with 100% test coverage if you like.

PS: You could also argue whether LRU is actually the best caching strategy. FIFO might be much faster for the majority of cases. But that's a different story.

wraithgar:
> I'm happy to provide a PR with 100% test coverage if you like.

On one hand it does seem like a simple Map cache is all we need and should be easy to implement.

On the other hand "why don't you just" and "how hard can it be?" are usually hard lessons once one goes down the path of answering them.

If it's not too much trouble, you can make that PR.  I think that will inform a lot of our decision better than an academic discussion would.

H4ad:
Closing in favor of #697

PR #696, opened 2024-04-07
chore: add benchmarks

H4ad:
I had these benchmarks from the last time I did some perf research in this repo, so I thought maybe it was a good idea to have them in the repo itself, so other people can use them and you guys can verify if the perf is still good or you had any regression.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ tap-testdir*/
 !/.gitignore
 !/.npmrc
 !/.release-please-manifest.json
+!/benchmarks
 !/bin/
 !/CHANGELOG*
 !/classes/
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   "devDependencies": {
     "@npmcli/eslint-config": "^4.0.0",
     "@npmcli/template-oss": "4.21.3",
+    "benchmark": "^2.1.4",
     "tap": "^16.0.0"
   },
   "license": "ISC",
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -37,3 +37,4 @@ tap-testdir*/
 !/tap-snapshots/
 !/test/
 !/tsconfig.json
+!/benchmarks
```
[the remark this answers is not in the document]
wraithgar: I believe `templateOSS.allowPaths` is the package.json entry that affects this.

benchmarks/package.json
```diff
@@ -0,0 +1,13 @@
+{
```
wraithgar, line 1: Adding this feels way more cumbersome than just adding `benchmark` as a devDependency.

package.json
```diff
@@ -71,7 +72,8 @@
       "/ranges/",
       "/index.js",
       "/preload.js",
-      "/range.bnf"
+      "/range.bnf",
+      "/benchmarks"
```
wraithgar, line 76: `npm run template-oss-apply` will apply this and rebuild the .gitignore.

benchmarks/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
wraithgar, line 1: Don't need this file or the empty README.md

package.json
H4ad: I ran but I removed the changes on `.github/workflows` about erasing a folder called `release`.

wraithgar: your local version may have been out of date. looks like the linting is green now which means we should be good now.

wraithgar (approved):

PR #448, opened 2022-04-08
Revert breaking changes in 7.3.6

lukekarrys:
Fixes: #447

/cc @ljharb

Issue #447, opened 2022-04-07
Why new minor version 7.3.6 is incompitable with node 13.14.0


EliseevNP:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

error semver@7.3.6: The engine "node" is incompatible with this module. Expected version "^10.0.0 || ^12.0.0 || ^14.0.0 || >=16.0.0". Got "13.14.0"

### Expected Behavior

It's should be compitable or new version should be 8.0.0

### Steps To Reproduce

1. In this environment...
2. With this config...
3. Run '...'
4. See error...

### Environment

- yarn: v1.22.4
- npm:
- Node: 13.14.0
- OS:
- platform:

kburton:
Additionally, `lru-cache ^7.4.0` is not compatible with node 10.

ljharb:
Indeed; v7.0.0 had no engines declaration at all, v7.1.0 set it to `>= 10`, and v7.3.6 was a breaking change both for changing `engines.node` and for upgrading `lru-cache` to v7.

ireinhart:
Same here... is https://github.com/npm/node-semver/pull/434 the issue?

ljharb:
@ireinhart i don't think so; #442 and #433 certainly are though.

pjohnmeyer:
@archm49 I'm not sure why I was @-ed into this so I'm unsubscribing; you'll need to @-me again w/ some context if I'm expected to participate

manjeetbrar91:
facing same issue. or production deployment stuck due this issue. please provide solution.

pbadenski:
Sure this shouldn’t be a patch version change, but if you’re still on a non-maintained odd-numbered version 1.5 years after EOL you should really revisit your security processes...

PR #697, opened 2024-04-11
replace `lru-cache` dependency

mbtools:
This change removes the dependency on the `lru-cache` package. It is replaced by a comparable, but much simpler class in this package. The cache is still set at max 1000 entries and works the same way as before.

Performance looks to be very similar. However, it needs a realistic test case that pushes the limits of the cache. I will leave it to more experts to judge. 

## References

#447, #448
#695 
https://github.com/npm/cli/issues/7350


wraithgar:
Normally we try not to ask too much of contributors in the way of commits, and this is one of the reasons why we typically don't accept dependency change PRs.  For those we'd prefer if the removal of `lru-cache` was a discrete `deps: remove lru-cache` commit so that it shows up in the changelog.

If it's not too much to ask can you make that `package.json` change a single commit w/ an appropriate commit message?  Everything else you can keep in a second commit with a `fix:` prefix.

wraithgar:
Our PR linter will pass if the PR subject *or* every commit has a valid conventional commit message. I've purposely made the PR subject a non-conventional one to reflect the fact that we intend to rebase-merge this PR with the `deps` and `fix` commits.

classes/lrucache.js
```diff
@@ -0,0 +1,71 @@
+class LRUCache {
+  constructor (max = 0) {
```
wraithgar, line 2: If we are not exporting this, and we are not exposing a way for end users to change this value, there's no need for `max` to be anything other than a fixed value.

wraithgar: We also wouldn't need a type check that is only relevant in the unit test, and not in any real world use case.

classes/lrucache.js
```diff
@@ -0,0 +1,71 @@
+class LRUCache {
```
wraithgar, line 1: Should this live in `./internal`?

classes/lrucache.js
mbtools: done

classes/lrucache.js
mbtools: agree

wraithgar:
Sorry for the confusion on the commit messages. If this is more than you want to do go ahead and get it all in here however you want and then we can get this PR green and we'll branch off of it and get the commits right.

internal/lrucache.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  size () {
+    return this.map.size
+  }
+
+  * entries () {
+    for (const [key, value] of this.map) {
+      yield [key, value]
+    }
```
ljharb, line 67:
```suggestion
    return this.map.entries()
```
?

mbtools: I used a generator to make comparison with [lru-cache](https://github.com/isaacs/node-lru-cache/blob/main/tap-snapshots/test/basic.ts.test.cjs) easier (it's just in reverse order). 

We can certainly switch to a regular function and adjust the snapshots.

mbtools:
Back to fully working. I will let you create a new PR with two commits as required. Thx!

wraithgar:
Thank you!  We'll let this sit for a bit for the community to weigh in, and we need some time to properly review it ourselves (we are in the middle of a large refactor in npm itself).

internal/lrucache.js
```diff
@@ -0,0 +1,69 @@
+class LRUCache {
+  constructor (max) {
+    if (!Number.isInteger(max) || max < 0) {
+      throw new TypeError('max must be a nonnegative integer')
+    }
```
lukekarrys, line 5: Now that this is an internal cache I think we can remove this check.

lukekarrys (commented):
I'm in favor of this PR and converting our `lru-cache` to an internal dependency.

If we do land this, I think we should remove any functionality that we aren't currently using in `semver`. Looking at the code that is only `cache.set()` and `cache.get()` (and `set` will call `delete()`). All my specific comments are about removing things. This is just my opinion and will let others weigh in first. So no need for anyone to make all these changes immediately 😄

test/internal/lrucache.js
```diff
@@ -0,0 +1,112 @@
...
+  t.equal(c.size(), 10)
+  t.matchSnapshot(c.entries())
+
+  // 9. Unset an item
+  t.equal(c.has(0), true)
+  t.equal(c.size(), 10)
+  c.set(1, undefined)
+  t.equal(c.has(1), false)
+  t.equal(c.size(), 9)
+  t.matchSnapshot(c.entries())
```
lukekarrys, line 87: I don't think the snapshot tests are particularly helpful here. For an implementation like this I think only testing the functionality is sufficient.

wraithgar:
I'm in favor of the same @lukekarrys.  I would like to see us get to a point where we're not testing semver's internal cache directly, but only as a side effect.  That is, only testing it exactly how it's used during normal semver operations.  Unit testing the module itself isn't that useful, all it does is show we HAVE a working cache, not that it's used by semver properly.

wraithgar:
@lukekarrys the title of this PR is intentionally not conventional so that we remember to rebase it into two commits: one for the deps change and one for the feat

mbtools:
Sounds good. We can certainly reduce the cache functions to a minimum. I just had them in there to prove to myself that it's working like lru-cache. 

If we're convinced the cache is fine, I suggest removing `entries()`, `capacity()`, the snapshots, and the `max type check`. But keep `has()` and `size()` so that the basic operations test still completes and we have 100% coverage. 

The cache is transparent to the outside of range parse. Showing that semver uses it properly, is / should be visible only using runtime measurements. Not sure how you prefer to implement those.

wraithgar:
> The cache is transparent to the outside of range parse. Showing that semver uses it properly, is / should be visible only using runtime measurements. Not sure how you prefer to implement those.

Here's a proof of concept showing how we could assure that subsequent `new Range` instances get back the same range set.

```javascript
const Semver = require('.')
const cached = Symbol('cached')
const r1 = new Semver.Range('1.0.0')
r1.set[0][cached] = true
const r2 = new Semver.Range('1.0.0')
console.log(r1.set[0][cached]) // Will be true
console.log(r2.set[0][cached]) // Will be true, showing it's cached.
```

mbtools:
- Removed unnecessary cache methods & tests
- Kept basic cache test to ensure 100% coverage
- Added a test to range class as suggested

wraithgar:
> Kept basic cache test to ensure 100% coverage

This still runs the risk that we have code in there we never use and don't need.  If we aren't using the code from `range` we don't need to test it.

I think the cache size is the only thing left right?  Can we loop through 1001 versions to test that the first one dropped out of the cache?  Can the max size be hard set in lru-cache for now?  There's no need to pass it in if we never set it to anything else.

lukekarrys:
@wraithgar We still use a coverage map in this repo. I'm think that will require testing the cache directly to get full coverage, right?

wraithgar:
We can edit that file.

wraithgar:
Here's an example of how we did it in npm itself when we wanted to test a component of all the commands in a new test file:

https://github.com/npm/cli/blob/21b823edf7b3095ce1f53da7befda2b41296afe4/test/coverage-map.js

wraithgar:
I think that we should fix the tests but also not block this PR on "perfect" code.  I'm good w/ this as-is (pending an actual review of the code at this point).  We can clean it up on our own time.

wraithgar:
We have not forgotten about this PR.  The big npm display layer refactor is almost done and then we can come back to things like this PR.

wraithgar:
https://github.com/npm/node-semver/pull/704

PR #713, opened 2024-05-08
fix(lru): use map.delete() directly

negezor:
The default implementation already returns boolean if the value has been deleted.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete#return_value

It's also faster since we don't double check the hashmap for a value.

### DIFF
--- a/internal/lrucache.js
+++ b/internal/lrucache.js
@@ -17,7 +17,7 @@ class LRUCache {
   }
 
   delete (key) {
-    return this.map.delete(key);
+    return this.map.delete(key)
   }
 
   set (key, value) {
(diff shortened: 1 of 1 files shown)


internal/lrucache.js
```diff
@@ -17,12 +17,7 @@ class LRUCache {
   }
 
   delete (key) {
-    if (this.map.has(key)) {
-      this.map.delete(key)
-      return true
-    } else {
-      return false
-    }
+    return this.map.delete(key);
```
lukekarrys, line 20:
```suggestion
    return this.map.delete(key)
```

lukekarrys (approved):
Thank you!

wraithgar (approved):

mbtools:
Thanks for the optimization 😊

PR #725, opened 2024-07-06
fix(lru): don't clear an already cached key on set

jviide:
This pull request modifies how the LRU cache implementation handles setting a key that's already in the cache. Previously, if an already cached key was set again, then it was always cleared from the cache:

```js
> const cache = new LRUCache();
> cache.set(1, 1);
> cache.set(1, 2);
> cache.get(1);
undefined
```

This pull request's modified version returns `2` for `cache.get(1)`. The newly set key is still bumped to the last place in the LRU queue. There are two new tests to check this behavior.

The current semver implementation might never trigger this behavior, in which case this pull request is just future-proofing.

### DIFF
--- a/internal/lrucache.js
+++ b/internal/lrucache.js
@@ -21,9 +21,9 @@ class LRUCache {
   }
 
   set (key, value) {
-    const deleted = this.delete(key)
+    this.delete(key)
 
-    if (!deleted && value !== undefined) {
+    if (value !== undefined) {
       // If cache is full, delete the least recently used item
       if (this.map.size >= this.max) {
         const firstKey = this.map.keys().next().value
--- a/test/internal/lrucache.js
+++ b/test/internal/lrucache.js
@@ -17,3 +17,25 @@ test('basic cache operation', t => {
   c.set(42, undefined)
   t.end()
 })
+
+test('test setting the same key twice', t => {
+  const c = new LRUCache()
+  c.set(1, 1)
+  c.set(1, 2)
+  t.equal(c.get(1), 2)
+  t.end()
+})
+
+test('test that setting an already cached key bumps it last in the LRU queue', t => {
+  const c = new LRUCache()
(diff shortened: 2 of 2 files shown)


wraithgar:
This is a great example of practically dead code that unit tests could never catch.  The way that `classes/range.js` uses this lib is to call `get` and if it's in there return it.  If it's not, it creates a new one and `set`s it.  All of the code between creating the range and `set`ting it is synchronous, no other event loop action could happen.  It's currently impossible for a `set` to even be called on an entry that already exists.

Tests that only interacted with ranges would have shown that some of this code was unreachable from the surface area of the library.  Unfortunately those tests don't exist, and unless someone wanted to put in that work we have what we have today.

With all that in mind, I propose that the real optimization here is to remove the first `this.delete(key)` altogether.  It's a noop during the current actual use case of this lru code.

internal/lrucache.js
```diff
@@ -21,9 +21,9 @@ class LRUCache {
   }
 
   set (key, value) {
-    const deleted = this.delete(key)
+    this.delete(key)
```
wraithgar, line 25:
```suggestion

```

jviide: This pull request was not meant as an optimization, but a fix to the LRU semantics. The two added unit tests for the lrucache.js module test the behavior.

Removing the `this.delete(key)` would make the LRU cache implementation into a mixture of LRU and FIFO caches. It may not be relevant with how the module is _currently_ used, but calling it an LRU cache could be an issue if someone were to expect it to work like a typical one in the future.

As such I recommend either keeping the delete call, removing it and calling the cache implementation something else than an LRU, or just closing this pull request. From the performance perspective removing or keeping the delete call seems to make no difference in the benchmarks.

wraithgar:
Gonna err on the side of caution then and close this as not necessary. Thanks for the time spent, all the same.

PR #564, opened 2023-06-15
fix: better handling of whitespace

lukekarrys:

PR #726, opened 2024-07-07
fix: optimize Range parsing and formatting

jviide:
This pull request optimizes the Range class in the following ways:

 1. Produce fewer intermediate objects when reducing a range's space characters to single spaces. This seems to improve bench-subset scores by up to 5%, and bench-satisfies scores to a lesser degree.
 2. Optimize Range formatting with explicit for loops instead, avoiding an intermediate array creation. This seems to improve bench-satisfies and bench-subset scores by up to 20%.
 3. Calculate Range's `.range` string (used by `.format()` and `.toString()`) lazily. This seems to improve bench-satisfies and bench-subset scores by up to 9%.

The external interface for the class stays the same, except for the new internal `.formatted` property used to cache its lazily calculated string. `Range#range` property is now also read-only.

There is a new test lazy formatting to ensure full test coverage.

The benchmarks bench-satisfies and bench-subset benefit from these changes, sometimes by up to 40%. Other benchmark results seem to stay the same. Here are the affected benchmarks before:

```
$ node benchmarks/bench-satisfies.js
satisfies(1.0.6, 1.0.3||^2.0.0) x 695,094 ops/sec ±0.68% (97 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0) x 764,115 ops/sec ±0.40% (99 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0) x 805,593 ops/sec ±0.62% (97 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true}) x 695,045 ops/sec ±0.73% (95 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true}) x 750,433 ops/sec ±0.66% (99 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true}) x 787,903 ops/sec ±0.39% (99 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true}) x 652,166 ops/sec ±0.34% (99 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true}) x 696,377 ops/sec ±0.36% (96 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true}) x 721,729 ops/sec ±0.35% (98 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 585,692 ops/sec ±0.75% (95 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 631,653 ops/sec ±0.33% (96 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 650,110 ops/sec ±0.64% (95 runs sampled)

$ node benchmarks/bench-subset.js
subset(1.2.3, *) x 633,342 ops/sec ±0.61% (95 runs sampled)
subset(^1.2.3, *) x 743,036 ops/sec ±0.47% (97 runs sampled)
subset(^1.2.3-pre.0, *) x 680,087 ops/sec ±0.76% (98 runs sampled)
subset(^1.2.3-pre.0, *) x 680,948 ops/sec ±0.46% (96 runs sampled)
subset(1 || 2 || 3, *) x 330,669 ops/sec ±0.53% (98 runs sampled)
```

And after:

```
$ node benchmarks/bench-satisfies.js
satisfies(1.0.6, 1.0.3||^2.0.0) x 896,936 ops/sec ±0.53% (94 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0) x 998,214 ops/sec ±0.40% (95 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0) x 1,000,593 ops/sec ±0.43% (97 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true}) x 890,369 ops/sec ±0.41% (100 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true}) x 977,239 ops/sec ±0.48% (97 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true}) x 983,682 ops/sec ±0.95% (96 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true}) x 805,330 ops/sec ±0.84% (98 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true}) x 894,117 ops/sec ±0.43% (99 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true}) x 911,742 ops/sec ±0.42% (96 runs sampled)
satisfies(1.0.6, 1.0.3||^2.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 741,254 ops/sec ±0.35% (97 runs sampled)
satisfies(1.0.6, 2.2.2||~3.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 807,380 ops/sec ±0.42% (99 runs sampled)
satisfies(1.0.6, 2.3.0||<4.0.0, {"includePrelease":true,"loose":true,"rtl":true}) x 820,390 ops/sec ±0.37% (99 runs sampled)

$ node benchmarks/bench-subset.js
subset(1.2.3, *) x 905,030 ops/sec ±0.63% (96 runs sampled)
subset(^1.2.3, *) x 1,026,457 ops/sec ±0.63% (95 runs sampled)
subset(^1.2.3-pre.0, *) x 923,789 ops/sec ±0.41% (97 runs sampled)
subset(^1.2.3-pre.0, *) x 923,136 ops/sec ±0.44% (96 runs sampled)
subset(1 || 2 || 3, *) x 432,037 ops/sec ±0.67% (96 runs sampled)
```

### DIFF
--- a/classes/range.js
+++ b/classes/range.js
@@ -1,3 +1,5 @@
+const SPACE_CHARACTERS = /\s+/g
+
 // hoisted class for cyclic dependency
 class Range {
   constructor (range, options) {
@@ -29,10 +31,7 @@ class Range {
     // First reduce all whitespace as much as possible so we do not have to rely
     // on potentially slow regexes like \s*. This is then stored and used for
     // future error messages as well.
-    this.raw = range
-      .trim()
-      .split(/\s+/)
-      .join(' ')
+    this.raw = range.trim().replace(SPACE_CHARACTERS, ' ')
 
     // First, split on ||
     this.set = this.raw
--- a/classes/range.js
+++ b/classes/range.js
@@ -20,7 +20,7 @@ class Range {
       // just put it in the set and return
       this.raw = range.value
       this.set = [[range]]
-      this.format()
+      this.formatted = undefined
       return this
     }
 
@@ -65,23 +65,29 @@ class Range {
       }
     }
 
-    this.format()
+    this.formatted = undefined
+  }
+
+  get range () {
+    if (this.formatted === undefined) {
+      this.formatted = ''
+      for (let i = 0; i < this.set.length; i++) {
+        if (i > 0) {
+          this.formatted += '||'
+        }
+        const comps = this.set[i]
+        for (let k = 0; k < comps.length; k++) {
+          if (k > 0) {
+            this.formatted += ' '
+          }
+          this.formatted += comps[k].toString().trim()
+        }
+      }
+    }
+    return this.formatted
   }
 
   format () {
-    this.range = ''
-    for (let i = 0; i < this.set.length; i++) {
-      if (i > 0) {
-        this.range += '||'
-      }
-      const comps = this.set[i]
-      for (let k = 0; k < comps.length; k++) {
-        if (k > 0) {
-          this.range += ' '
-        }
-        this.range += comps[k].toString().trim()
-      }
-    }
     return this.range
   }
 
--- a/classes/range.js
+++ b/classes/range.js
@@ -69,10 +69,19 @@ class Range {
   }
 
   format () {
-    this.range = this.set
-      .map((comps) => comps.join(' ').trim())
-      .join('||')
-      .trim()
+    this.range = ''
+    for (let i = 0; i < this.set.length; i++) {
+      if (i > 0) {
+        this.range += '||'
+      }
+      const comps = this.set[i]
+      for (let k = 0; k < comps.length; k++) {
+        if (k > 0) {
+          this.range += ' '
+        }
+        this.range += comps[k].toString().trim()
+      }
+    }
     return this.range
   }
 
--- a/test/classes/range.js
+++ b/test/classes/range.js
@@ -82,6 +82,15 @@ test('tostrings', (t) => {
   t.end()
 })
 
+test('formatted value is calculated lazily and cached', (t) => {
+  const r = new Range('>= v1.2.3')
+  t.equal(r.formatted, undefined)
+  t.equal(r.format(), '>=1.2.3')
+  t.equal(r.formatted, '>=1.2.3')
+  t.equal(r.format(), '>=1.2.3')
+  t.end()
+})
+
 test('ranges intersect', (t) => {
   rangeIntersection.forEach(([r0, r1, expect]) => {
     t.test(`${r0} <~> ${r1}`, t => {


classes/range.js
```diff
@@ -29,10 +29,11 @@ class Range {
     // First reduce all whitespace as much as possible so we do not have to rely
     // on potentially slow regexes like \s*. This is then stored and used for
     // future error messages as well.
```
kurtextrem, line 31: probably worth looking up why the comment has made it into the source (was there a bench or sth?), and if the reason no longer makes sense today, probably worth removing.

classes/range.js
```diff
@@ -29,10 +29,11 @@ class Range {
...
-    this.raw = range
-      .trim()
-      .split(/\s+/)
-      .join(' ')
+    this.raw = range.replace(/\s+/g, ' ')
```
kurtextrem, line 32: does it make sense to hoist the regex?

classes/range.js
```diff
@@ -66,14 +67,19 @@ class Range {
...
 
-    this.format()
+    this.formatted = undefined
+  }
+
+  get range () {
+    if (this.formatted === undefined) {
+      this.formatted = this.set
+        .map((comps) => comps.join(' ').trim())
+        .join('||')
```
kurtextrem, line 77: How often is `range` usually called, would it make sense to replace this with a for loop too?

classes/range.js
jviide: The code's functionality has not really changed since the comment was added in #564 by @lukekarrys. As such I'm not able to comment. However, I think investigating it falls somewhat outside the scope of this pull request.

classes/range.js
jviide: I did benchmark a the hoisted version prior to submitting this pull request, and it did not have a measurable impact in my environment. Hoisting won't hurt here, though, I'll do that 👍

classes/range.js
kurtextrem: My intention was rather to figure out if we should remove it, because it says "...so we do not have to rely on potentially slow regexes like \s*". In this PR we switch to a regex, so the comment feels slightly out of place.

classes/range.js
jviide: That's a good question! The benchmarks don't seem to contain anything that get the `.range` property. Transforming this into loops like this seems to speed up the calculation considerably though:

```js
  get range () {
    if (this.formatted === undefined) {
      let formatted = ''
      for (let i = 0; i < this.set.length; i++) {
        if (i > 0) {
          formatted += '||'
        }
        const comps = this.set[i]
        for (let k = 0; k < comps.length; k++) {
          if (k > 0) {
            formatted += ' '
          }
          formatted += comps[k].toString().trim()
        }
      }
      this.formatted = formatted
    }
    return this.formatted
  }
```

classes/range.js
jviide: The previous version also used a similar regex:

```js
this.raw = range
  .trim()
  .split(/\s+/)
  .join(' ')
```

It was introduced in the same pull request as the comment: https://github.com/npm/node-semver/pull/564/files#diff-95bd101ea814ee8011aedabd5c1d1b1a468dbfbf774168e3c3e319a332deadb6R29-R35

jviide: Ah, I think I got it! The same pull request introduces this change to the regular expressions used elsewhere in the library: https://github.com/npm/node-semver/pull/564/files#diff-7e884a3dd192e2887678190dd2acabc002cd4ecb2e7b5801d33ffa5248b98302R12-R28

That's why the whitespaces need to get collapsed, and why `\s*` regexps can be later avoided.

jviide: As such I think the comment still holds, as it's not referring to the `\s+` regexp used immediate below it (?)

classes/range.js
jviide: I added this in a new commit that's applied before the lazy calculation commit. By itself this change seemed to boost bench-satisfies and bench-subset scores by up to 20%. Thanks for the suggestion! 👍

kurtextrem: Nice win!

jviide:
The "refactor:" pull request name prefix was not allowed so I picked "fix:", and changed the commit messages to use the same prefix as well.

wraithgar:
We generally squash-merge external PRs unless there's a compelling reason to keep each commit.  linting will pass if *either* the PR title or every commit has a valid conventional prefix.

Don't worry about keeping the commit messages "valid" as we go through this PR, as we'll be using the PR title if and when this lands.

classes/range.js
```diff
@@ -1,3 +1,5 @@
+const SPACE_CHARACTERS = /\s+/g
```
wraithgar, line 1: This is only used once, there's no need to abstract it to a variable.

functions/inc.js
```diff
@@ -9,7 +9,7 @@ const inc = (version, release, options, identifier, identifierBase) => {
 
   try {
     return new SemVer(
-      version instanceof SemVer ? version.version : version,
+      version instanceof SemVer ? version.raw : version,
```
wraithgar, line 12: This change seems unrelated to the rest of the PR and may represent a subtly breaking change.  At the very least it needs to be a separate PR where its merits can be explored.

test/classes/range.js
```diff
@@ -82,6 +82,13 @@ test('tostrings', (t) => {
   t.end()
 })
 
+test('format', (t) => {
```
wraithgar, line 85: Future devs will probably appreciate a more descriptive test here.  We know today it's for coverage of the `this.formatted` test but that's not apparent from just this test alone.

wraithgar:
Thanks for the help in reviewing this @kurtextrem.  Yes, that comment was referring to `s*` which is a classic ReDOS vector.  Historically we've used the `.split` approach to fix that, because it's rarely in a hot path where performance matters as much as it does here.   Good to know that `String.replace` is faster.

classes/range.js
marvinhagemeister: FYI: This is a performance optimization, not a stylistic choice. Reconstructing a new `Regex` instance each time the `Range` class is instantiated is much more expensive then reusing the same one.

classes/semver.js
```diff
@@ -73,14 +73,20 @@ class SemVer {
     }
 
     this.build = m[5] ? m[5].split('.') : []
-    this.format()
+    this.formatted = undefined
```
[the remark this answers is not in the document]
marvinhagemeister: This is not a stylistic choice, but a performance optimization. By declaring all properties upfront, even if they are initialized to `undefined` you can avoid a shape transition in the JIT. That's what's done here.would make the code

test/classes/range.js
marvinhagemeister: To me this seems fine. The test is all about `.format()` calls.

functions/inc.js
jviide: This change was originally described as follows:

> a modification to inc to use the SemVer#raw property instead of SemVer#version to avoid extra formatting work when possible

As such this was a part of the original optimizations for the SemVer class. When a new SemVer object was created based on an older SemVer object, it forced the older SemVer instance to calculate its the `.version` property. To avoid this using the older SemVer instance's `.raw` property (which is the original string used to initialize the older SemVer instance) seemed reasonable.

I reverted this change, which caused the test "increment versions test" to start failing, specifically the "didn't modify its input" part. This happens because `version instanceof SemVer ? version.version : version` line can force the new `.formatted` property to change, technically modifying the SemVer instance used as the input.

Time constraints don't allow me to dive into the tests with confidence, so I removed the SemVer class related optimizations from this pull request.

test/classes/range.js
jviide: Renamed the test to "formatted value is calculated lazily and cached" and modified it to examine the `.formatted` property more directly.

classes/semver.js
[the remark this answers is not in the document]
jviide: The same reasoning applies for the similar changes in range.js (as this PR doesn't modify semver.js anymore).

wraithgar (approved):

PR #745, opened 2024-09-06
fix: Ignore uppercase and lowercase in clear

li-jia-nan:
## References

Ignore uppercase and lowercase in clear


wraithgar:
What is this supposed to be fixing?  What's the use case here?  A test would be needed at a minimum.

li-jia-nan:
> What is this supposed to be fixing? What's the use case here? A test would be needed at a minimum.

Hi, This PR fixes the problem of not supporting `V1.2.3` format strings. I added a test case, and I think the capital `V` should also be cleaned up.

PR #749, opened 2024-11-28
docs: clarify comment about obsolete prefixes

mbtools:
Adds an explanation why the "=" and "v" prefixes should not be used anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ A "version" is described by the `v2.0.0` specification found at
 <https://semver.org/>.
 
 A leading `"="` or `"v"` character is stripped off and ignored.
-These prefixes are kept for compatibility with `v1.0.0` of the SemVer
+A leading "v" is kept for compatibility with `v1.0.0` of the SemVer
 specification but should not be used anymore.
 
 ## Ranges
--- a/README.md
+++ b/README.md
@@ -141,6 +141,8 @@ A "version" is described by the `v2.0.0` specification found at
 <https://semver.org/>.
 
 A leading `"="` or `"v"` character is stripped off and ignored.
+These prefixes are kept for compatibility with `v1.0.0` of the SemVer
+specification but should not be used anymore.
 
 ## Ranges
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -141,6 +141,8 @@ A "version" is described by the `v2.0.0` specification found at
 <https://semver.org/>.
 
 A leading `"="` or `"v"` character is stripped off and ignored.
+These prefixes are kept for compatibility with `v1.0.0` of the SemVer
+specification but should not be used anymore.
```
ljharb, line 145: um, why? the = in a dependency is something i constantly use to be an explicit indicator that the pinning I’m doing is intentional.

(also i don’t think semver v1 had any range specifiers, including =)

mbtools: I get why `=1.2.3` makes sense visually although `1.2.3` does the same trick. 

How about `A leading "v" is kept for compatibility ...`?

ljharb: That sounds fine with me :-)

also i don't think the leading v is actually allowed by the semver spec - v1 requires it [for a git tag](https://semver.org/spec/v1.0.0.html#spec-item-11), and v2 just says [it's not allowed](https://semver.org/spec/v2.0.0.html#is-v123-a-semantic-version).

mbtools: Using v-tags (tags in general) is confusing since they must not necessarily match the package version.  Thankfully, it's pretty rare in the wild.

ljharb: in package.json yes. however, the git tags for published versions should and do start with a `v` :-)

wraithgar:
This now says two contradictory things.

 - `v` is stripped
 - `v` is kept

README.md
```diff
@@ -141,6 +141,8 @@ A "version" is described by the `v2.0.0` specification found at
 <https://semver.org/>.
 
 A leading `"="` or `"v"` character is stripped off and ignored.
+A leading "v" is kept for compatibility with `v1.0.0` of the SemVer
```
ljharb, line 144:
```suggestion
Support for stripping a leading "v" is kept for compatibility with `v1.0.0` of the SemVer
```

mbtools: Long day :-) Thanks.

wraithgar:
Gonna land this but won't publish a new version until another change besides this one comes along.

wraithgar (approved):

Issue #751, opened 2024-12-11
[BUG] inc premajor, preminor, and prepatch do the wrong thing if the version is already a pre-release.


wraithgar:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

From the docs:

> premajor in one call will bump the version up to the next major version and down to a prerelease of that major version. preminor, and prepatch work the same way.

If we do these steps in isolation (using the internal, "pre" that does the `down to a prerelease`) we get:

```javascript
> semver.inc('2.0.0-pre', 'major')
'2.0.0'
> semver.inc('2.0.0', 'pre')
'2.0.0-0'
```

But if we do `premajor`

```javascript
> semver.inc('2.0.0-pre', 'premajor')
'3.0.0-0'
```

### Expected Behavior

`premajor` et al should "do what it says on the tin", and not jump an extra major/minor/patch.

### Environment

`semver@7.6.3`

reggi:
real world use case: https://github.com/npm/template-oss/pull/495

ljharb:
presumably 2.0.0 + premajor would still jump to 3.0.0-0?

wraithgar:
Yes, they all currently do the right thing if the version being incremented is not already a prerelease.

```javascript
> semver.inc('2.0.0', 'major')
'3.0.0'
> semver.inc('3.0.0', 'pre')
'3.0.0-0'
> semver.inc('2.0.0', 'premajor')
'3.0.0-0'
```

wraithgar:
```
> semver.inc('3.2.4-pre.1', 'premajor')
'4.0.0-0'
```

It seems we also drop the existing preid, unlike

```
> semver.inc('3.2.4-pre.1', 'prerelease')
'3.2.4-pre.2'
```

reggi:
I also want to note that semver inc currently isn't always strict and does "just drop prerelease" 

Incremented "premajor version" will always drop prerelease.

```
> semver.inc('7.0.0-pre.0', 'major')
'7.0.0'
> semver.inc('7.0.0-pre.0', 'minor')
'7.0.0'
> semver.inc('7.0.0-pre.0', 'patch')
'7.0.0'
```

Incremented "preminor version" will drop prerelease if minor change.

```
> semver.inc('6.1.0-pre.0', 'major')
'7.0.0'
> semver.inc('6.1.0-pre.0', 'minor')
'6.1.0'
> semver.inc('6.1.0-pre.0', 'patch')
'6.1.0'
```

[...]

Issue #740, opened 2024-08-20
[ENHANCMENT] Add explicit "release" identifier to `inc` function to strip prerelease suffix from the version.


deej-io:
I have been building a release process for one of my projects using npm-version and wanted to promote a prerelease version to a full version and it wasn't immediately obvious how to do that.

After delving into the code, I found that `patch` effectively does what I want, however I think this is confusing and potentially error-prone when running the command in a context where we don't know whether the current version is a pre-release or not - e.g. in a CI job.

I think it would be beneficial to have a `release` increment identifier that strips the prerelease suffix from the version. This is different from the `patch` identifier in that it does not bump the patch version in the case where there is no prerelease suffix.

Instead, this identifier could either:

* Emit an error that the the current version is not a pre-release; or
* Return the version unmodified

I would personally prefer it to return an error, however I am open to reasons why a no-op would be better.

I am happy to implement this and propagate the change up to npm-version.

wraithgar:
This seems very discrete and user-friendly.  So the idea is that npm could then support `npm version release`?

One of the biggest lessons I've learned in maintaining npm is that it's always best to be explicit and never guess user intent.  With that in mind I'd lean towards throwing if the version does not have a pre-release portion.

mbtools:
Like this? 

```js
      case 'release':
        if (this.prerelease.length) {
          throw new Error('version is not a pre-release')          
        }
        this.prerelease = []
        break
```

LGTM 👍

deej-io:
> This seems very discrete and user-friendly. So the idea is that npm could then support `npm version release`?

That's the overall intention, yeah.

> One of the biggest lessons I've learned in maintaining npm is that it's always best to be explicit and never guess user intent. With that in mind I'd lean towards throwing if the version does not have a pre-release portion.

Agreed! Thank you for the feedback :)

wraithgar:
`if (!this.prerelease.length) {`  I should hope, right?

mbtools:
yes, `!` of course 😄

wraithgar:
```javascript
> (new (require('.').SemVer)('1.0.0')).inc('release')
Uncaught Error: version 1.0.0 is not a prerelease
    at SemVer.inc (classes/semver.js:181:17)
> require('.').inc('1.0.0', 'release')
null
```

please remember that `semver.inc` catches errors, only the semver object will throw.

wraithgar:
Draft PR at https://github.com/npm/node-semver/pull/752

PR #752, opened 2024-12-11
fix pre* increments, add release increment

wraithgar:
This PR has two changes:

- **fix: don't leapfrog versions during premajor et al**
- **feat: add "release" inc type**

The bug is outlined in https://github.com/npm/node-semver/issues/751
The feature is defined in https://github.com/npm/node-semver/issues/740

This is a draft while we discuss #751.
This will need docs and tests of course.

### DIFF
--- a/classes/semver.js
+++ b/classes/semver.js
@@ -173,40 +173,31 @@ class SemVer {
     } while (++i)
   }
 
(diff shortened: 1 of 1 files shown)


wraithgar:
#753 implements `release`

PR #753, opened 2025-01-21
feat: add "release" inc type

### LINKED ISSUE #740: [ENHANCMENT] Add explicit "release" identifier to `inc` function to strip prerelease suffix from the version.

mbtools:
Here's the `release` options for `inc`. Includes update to docs and a few tests.

Closes #740

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Options:
 -i --increment [<level>]
         Increment a version by the specified level.  Level can
         be one of: major, minor, patch, premajor, preminor,
-        prepatch, or prerelease.  Default level is 'patch'.
+        prepatch, prerelease, or release.  Default level is 'patch'.
         Only one version may be specified.
 
 --preid <identifier>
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -100,7 +100,7 @@ Options:
 -i --increment [<level>]
         Increment a version by the specified level.  Level can
         be one of: major, minor, patch, premajor, preminor,
-        prepatch, or prerelease.  Default level is 'patch'.
+        prepatch, prerelease, or release.  Default level is 'patch'.
```
wraithgar, line 103: `inc()` is also documented under `#Functions` below and this should be added there.

wraithgar (approved):

PR #755, opened 2025-01-21
fix(diff): fix prerelease to stable version diff logic

### LINKED ISSUE #606: [bug] diff("1.7.2-1", "1.8.1") returns patch, not minor

### Current Behavior

`diff("1.7.2-1", "1.8.1")` returns "patch".

### Expected Behavior

I would have expected it to return "minor" since we're moving from 1.7 to 1.8.

### Steps To Reproduce

1. In this environment...
2. With this config...
3. Run '...'
4. See error...

### Environment

- npm:
- Node:
- OS:
- platform:

eminberkayd:
Refactored version diff logic to handle transitions from prerelease versions to stable versions correctly for major/minor/patch version bumps.

Closes #606

### DIFF
--- a/functions/diff.js
+++ b/functions/diff.js
@@ -29,14 +29,14 @@ const diff = (version1, version2) => {
 
     // Otherwise it can be determined by checking the high version
 
-    if (highVersion.patch) {
-      // anything higher than a patch bump would result in the wrong version
-      return 'patch'
-    }
-
     if (highVersion.minor) {
       // anything higher than a minor bump would result in the wrong version
       return 'minor'
+    }
+
+    if (highVersion.patch) {
+      // anything higher than a patch bump would result in the wrong version
+      return 'patch'
     }
 
     // bumping major/minor/patch all have same result
--- a/functions/diff.js
+++ b/functions/diff.js
@@ -27,20 +27,13 @@ const diff = (version1, version2) => {
       return 'major'
     }
 
-    // Otherwise it can be determined by checking the high version
-
-    if (highVersion.minor) {
-      // anything higher than a minor bump would result in the wrong version
-      return 'minor'
-    }
-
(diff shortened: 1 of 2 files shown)


wraithgar:
Looking at what needed to be done to fix this, and seeing the code that was left, I had a thought that this wasn't complete and I was right.

```sh
~/D/n/n/b/main (fix/prerelease-diff-logic-#606|✔) $ node
Welcome to Node.js v22.13.1.
Type ".help" for more information.
> require('.').diff('1.1.1-pre', '2.1.1')
'minor'
> 
```

So, I think you've identified what's wrong here (order of operations) but we obviously still have holes.

wraithgar:
```patch
diff --git a/functions/diff.js b/functions/diff.js
index f011d99..58f9fde 100644
--- a/functions/diff.js
+++ b/functions/diff.js
@@ -27,8 +27,13 @@ const diff = (version1, version2) => {
       return 'major'
     }
 
-    // Otherwise it can be determined by checking the high version
+    // If the majors differ then it's major
+    if (highVersion.major !== lowVersion.major) {
+      return 'major'
+    }
 
+    // Otherwise it can be determined by checking the high version
+    //
     if (highVersion.minor) {
       // anything higher than a minor bump would result in the wrong version
       return 'minor'
```

```sh
> require('.').diff('1.1.1-pre', '2.1.1-pre')
'premajor'
> require('.').diff('1.1.1-pre', '2.1.1')
'major'
> 
```

Tests still pass with this, would need to add test cases for this new case.

eminberkayd:
I have incorporated additional test cases and updated the PR accordingly. Kindly review it when you have time, and let me know if any modifications are required. Thank you!

wraithgar (approved):

wraithgar:
Oh yes this is much better.

Issue #763, opened 2025-02-01
[BUG] 7.6.0 --> 7.7.0 inc behavior change


danchurch-alphasights:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

We use [auto](https://github.com/intuit/auto) in our CI to manage our versioning.

For canary releases, Auto [uses a version format](https://github.com/intuit/auto/blob/c6961698ab27ac61b1a0f44a9e20785d5ae11e88/packages/core/src/auto.ts#L1304-L1310) of `<version>-canary.<pr#>.<hash>`

We started to see our CI randomly failing (based on commit hash or build hash) the last few days with the following error:
```
Error: Running command 'npx' with args [lerna, publish, prepatch, --dist-tag, canary, --registry, <REDACT>, --yes, --no-git-reset, --no-git-tag-version, --exact, --no-verify-access, --preid, canary.661.2207bf7] failed


Changes:
 - <REDACT>: 0.127.2 => null
 - <REDACT>: 0.35.0 => null

...

lerna ERR! Cannot read properties of null (reading 'trim')
lerna ERR! errno "undefined" is not a valid exit code - exiting with code 1
```

After digging deep into Lerna / Auto, I found out the root cause was due to a change of the semver package between 7.6.0 and 7.7.0

|       | `semver.inc('1.0.0', 'prepatch', 'canary.661.2207bf')` | `semver.valid('1.0.1-canary.661.2207bf.0')` |
| ----- | ----- | ----- |
| 7.6.0 | `1.0.1-canary.661.2207bf.0` | `1.0.1-canary.661.2207bf.0` |
| 7.7.0 | `null` | `1.0.1-canary.661.2207bf.0` |

### Expected Behavior

I would expect `semver.inc('1.0.0', 'prepatch', 'canary.661.2207bf')` to continue to return a valid value or `semver.valid('1.0.1-canary.661.2207bf.0')` to return `null`

### Steps To Reproduce

```bash
npm i semver@7.6.0 --save-exact

node -e "console.log(require('semver').inc('1.0.0', 'prepatch', 'canary.661.2207bf'));"
```

```bash
1.0.1-canary.661.2207bf.0
```

-------

```bash
npm i semver@7.7.0 --save-exact

node -e "console.log(require('semver').inc('1.0.0', 'prepatch', 'canary.661.2207bf'));"
```

```bash
null
```

-------

```bash
npm i semver@7.7.0 --save-exact

[...]

PR #764, opened 2025-02-03
fix(inc): fully capture prerelease identifier

wraithgar:
Closes: https://github.com/npm/node-semver/issues/763


wraithgar:
There is probably something deeper going on here that explains why the prerelease regex on its own doesn't capture the whole string, that may be worth looking into but if it's gonna take more than a day I think we ship this as-is and fix it later.

owlstronaut (approved):

PR #773, opened 2025-03-20
chore(package): explicitly declare js module type

Fdawgs:
[Node 21.1.0 added a flag to detect module types](https://github.com/nodejs/node/releases/tag/v21.1.0) that became enabled by [default in Node 22.7.0](https://nodejs.org/api/packages.html#syntax-detection).
Declaring the type will cause Node to skip detection on startup, reducing startup time.

Declaring the package type is also considered good practice according to https://nodejs.org/api/modules.html#enabling.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "7.7.1",
   "description": "The semantic version parser used by npm.",
   "main": "index.js",
+  "type": "commonjs",
   "scripts": {
     "test": "tap",
     "snap": "tap",


owlstronaut:
Hello @Fdawgs, thank you very much for taking the time to contribute and for your suggestion. After reviewing the change, we’ve determined that adding "type": "commonjs" isn’t necessary for our tooling. Since Node.js already treats .js files as CommonJS by default (when no "type" field is present) and because our performance and compatibility aren’t impacted in this case, incorporating this change would only add extra complexity to our toolchain without any clear benefit.

More on the matter here: https://nodejs.org/en/blog/release/v22.7.0#module-syntax-detection-is-now-enabled-by-default

> Syntax detection should have no performance impact on CommonJS modules

Fdawgs:
No worries, thanks for the response!

PR #781, opened 2025-04-24
fix: prerelease identifier starting with digits

### LINKED ISSUE #775: [BUG] Coercing version with prerelease identifier that starts with digits returns truncated identifier

### Current Behavior

Using semver `7.7.1` (and possibly earlier versions), attempting to coerce a version string that contains a prerelease identifier that starts with digits, will result in that identifier to be truncated after the digits.

Here are some example:

- `1.0.0-alpha.1ab` produces `1.0.0-alpha.1`
- `1.0.0-alpha.12ab` produces `1.0.0-alpha.12`
- `1.0.0-alpha.1234.23cd` produces `1.0.0-alpha.1234.23`

This is problematic when the identifier is a hash that may start with a digit.

The issue doesn't seem to happen if the identifier is composed of only number, or if the identifier starts with an alphabetic character:

- `1.9.5-nightly.abc123` is coerced as expected
- `1.9.5-nightly.abcdef` is coerced as expected
- `1.9.5-nightly.123456` is coerced as expected

### Expected Behavior

The prerelease identifier isn't truncated.

### Steps To Reproduce

1. Clone the reproduction repository: https://github.com/nhedger/semver-prerelease-issue
2. Install the dependencies
3. Run `node index.mjs`
4. See that the prerelease identifier is truncated

### Environment

- npm: 10.8.2
- Node: 20.17.0
- OS: macOS
- platform: MacBook Pro M1

mbtools:
Closes #775

@wraithgar your [suspicion](https://github.com/npm/node-semver/pull/764#issuecomment-2631653070) was correct. The regex for the prerelease identifier was not precise enough. Changing the order of numeric and non-numeric parts fixes the issue. I reverted the change to `inc` function in `semver.js`. The tests still pass 😃 

I added tests that check the exact result of parsing valid versions in particular the pre-release and build parts which was somehow missing. We might want to add edge cases here.


mbtools:
Looks like the lint errors are related to template, not this change

wraithgar:
Thanks, good move using the `re` that's already being exported instead of trying to make a new one.

wraithgar (approved):

PR #787, opened 2025-05-13
feat: moving the project to typescript

jarrodek:
- It moves the project as-is to typescript.
- It replaces the benchmark library with `@pawel-up/benchmark`


jarrodek:
Sorry, wrong upstream.

Issue #691, opened 2024-03-20
[BUG] limit version prefix to optional `v`


mbtools:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The following "ranges" are currently considered valid: > v=1.2.3, > vvv1.2.3, > v==1.2.3, > v=vv==v1.x

The validation is not strict enough and allows `=`, `v=`, and nonsensical prefixes as well.

### Expected Behavior

With next major semver, the version prefix should be limited to a single, optional `v`. Other prefixes shall not be permitted anymore.

Suggested change:

```js
// createToken('LOOSEPLAIN', `[v=\\s]*${src[t.MAINVERSIONLOOSE] // old
createToken('LOOSEPLAIN', `v?\\s*${src[t.MAINVERSIONLOOSE] // new

// createToken('XRANGEPLAIN', `[v=\\s]*(${src[t.XRANGEIDENTIFIER]})` +  // old
createToken('XRANGEPLAIN', `v?(${src[t.XRANGEIDENTIFIER]})` +// new

// createToken('XRANGEPLAINLOOSE', `[v=\\s]*(${src[t.XRANGEIDENTIFIERLOOSE]})` + // old
createToken('XRANGEPLAINLOOSE', `v?\\s*(${src[t.XRANGEIDENTIFIERLOOSE]})` + // new
```

### Steps To Reproduce

```js
validRange('> v=1.2.3') 
```

### Environment

- node-semver: 7.6.0

wraithgar:
Added this to https://github.com/npm/node-semver/issues/501

mbtools:
I have done some analysis of all versions, dependencies and engine ranges used in the top ~5000 npm `package.json` files:

Type of Range | Count
-- | --
^ | 52049
semver | 14049
\>= | 2934
~ | 1979
\|\| | 841
\* | 730
\x | 502
major | 348
\= | 162
"latest" | 154
npm: | 101
git: | 69
\- | 60
file: | 44
\> | 25
major.minor | 24
"next" | 15
http/s: | 12
link: | 7
portal: | 3
\< | 3
other: . | 3
other: v1.1.0 | 1
other: ./ | 1
workspace: | 1
other: v1.13.4 | 1
Grand Total | 74118

The use of `v` or `v=` prefixes is extremely rare. It's only 2 cases (which could be tags and valid).  

[...]

PR #789, opened 2025-06-23
feat: allow strict semver check

riderx:
The Current Implementation does not follow SemVer v2 as stated in the doc.
This allows following the spec by making leading v as invalid with an option named "strict".

I'm open to renaming it if needed.

## References
Related to https://github.com/npm/node-semver/issues/376
and https://github.com/npm/node-semver/issues/376
and PR: https://github.com/npm/node-semver/pull/476 who was rejected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,11 @@ semver.lt('1.2.3', '9.8.7') // true
 semver.minVersion('>=1.0.0') // '1.0.0'
 semver.valid(semver.coerce('v2')) // '2.0.0'
 semver.valid(semver.coerce('42.6.7.9.3-alpha')) // '42.6.7'
+
+// Strict mode - rejects version strings with 'v' prefix
+semver.valid('v1.2.3') // '1.2.3' (default behavior)
+semver.valid('v1.2.3', { strict: true }) // throws TypeError
+semver.valid('1.2.3', { strict: true }) // '1.2.3'
 ```
 
 You can also just load the module for the function that you care about if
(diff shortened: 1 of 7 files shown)


wraithgar:
This introduces a bit of a confusing paradigm here now.  We already have the `loose` option, which defaults to `true`.  I think adding a `strict` config that now solely interacts with this one part of parsing is confusing and incorrect.  Folks would assume (correctly I would say) that these configs are reflections of each other and should act as such.

This is a rather subtle and complex problem to solve unfortunately.  The `v`-prefixed version is likely here to say based on how disruptive removing it would be.

I think there is still merit in having a way to tell this package to ignore that vestigial support, but I don't think adding a `strict` option to the constructor is the full solution.  Also I unfortunately don't have any good input on what the full solution is, which I know is frustrating.  More discussion is needed here, and this PR is a good place to have it for now.

riderx:
@wraithgar thanks for the prompt reply, I see what you mean and you are right.
I think I could do a SemVer2 option and this will set the loose to false then.

Do you think it could less confusing and valid ?

ljharb:
I've discussed it with the spec maintainers and there's PRs open for much of what I discussed.

npm's flavor of semver is much more valuable and widespread than the spec itself - specs should document reality, not dictate it.

mbtools:
Close to no one is using the `v` prefix anyway (see [here](https://github.com/npm/node-semver/issues/691#issuecomment-2015823124)). Removing the prefix in next major won't be disruptive.

riderx:
It does not “ Example: git tag v1.2.3 -m "Release version 1.2.3", in which case “v1.2.3” is a tag name and the semantic version is “1.2.3”.” as you can see. And none of any others package i tried who implement the spec of Semver in others languages /platform allow it. 
for my experience, java, kotlin, swift, deno

riderx:
I agree i do see the prefix used in git tag a lot, but not in package itself. 
a real data who could help would be to find usage of the lib in code and see how many time it’s used to parse git tag, but it seems impossible mission to properly do it. Maybe @mbtools you know some magic for this?
To me the best way would be to release a sub package, like semver/strict (if we can get the stats otherwise a different name), and see if over time the stats lean to it with of course a clear incentive in the readme.
if that the case then we know that the direction the community expects to package to go.
Data would be better than any conversation to me :)

mbtools:
dear wizard of almighty ai, please scan 1/2 billion GH repos and 50 mill on gitlab for semver parsing a git tag

owlstronaut:
Closing as abandoned

PR #793, opened 2025-06-27
chore: add type to package json

UltiRequiem:
## What? 

I added the type property to package.json.

## Why 

Node.js may attempt to detect the package type causing a small performance hit.

https://nodejs.org/api/packages.html#determining-module-system

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "7.7.2",
   "description": "The semantic version parser used by npm.",
   "main": "index.js",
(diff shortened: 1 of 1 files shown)


wraithgar:
Attributes like this are centrally managed by [template-oss](https://github.com/npm/template-oss), which allows us to standardize on how we populate these kinds of things.

We do not have any plans right now to add this, but we may in the future.

UltiRequiem:
πππ

PR #854, opened 2026-04-29
fix: handle OR-range unions in subset() (#703)

### LINKED ISSUE #703: [BUG] `subset('>=17.2.0', '^17.2.0 || >17')` should be `true`

### Current Behavior

```js
const subset = require("semver/ranges/subset")

subset('^17.2.0 || >17', '>=17.2.0') // true
subset('>=17.2.0', '^17.2.0 || >17') // false

subset('^17.2.0 || >=18', '>=17.2.0') // true
subset('>=17.2.0', '^17.2.0 || >=18') // false
```

### Expected Behavior

```js
const subset = require("semver/ranges/subset")

subset('^17.2.0 || >17', '>=17.2.0') // true
subset('>=17.2.0', '^17.2.0 || >17') // true

subset('^17.2.0 || >=18', '>=17.2.0') // true
subset('>=17.2.0', '^17.2.0 || >=18') // true
```

### Steps To Reproduce

https://runkit.com/scagood/662900db83c1e60008f2ccd4

### Environment

- npm: 10.5.0
- Node: 20.12.2
- OS: Ubuntu 22.04.4
- platform: KDE

abhu85:
## Summary

Fixes #703

`subset('>=17.2.0', '^17.2.0 || >17')` incorrectly returns `false`.

The root cause is that `subset()` checks each OR-branch of the dom range independently. No single branch of `^17.2.0 || >17` covers `>=17.2.0` on its own, but their **union** does:

```
^17.2.0  →  [17.2.0, 18.0.0)
>17      →  [18.0.0, ∞)
union    →  [17.2.0, ∞)  ⊇  [17.2.0, ∞)  ✓
```

## Approach

When no single dom OR-branch covers a sub comparator set, fall back to an **interval-sweep** algorithm:

1. Extract `[lower, upper)` bounds from the sub and every dom OR-branch
2. Sort dom intervals by lower bound (ties broken by operator strictness)
3. Sweep left-to-right: start at sub's lower bound and greedily extend coverage through overlapping/adjacent dom intervals
4. If the sweep reaches or exceeds sub's upper bound, the union covers the sub

Key details:
- **Prerelease adjacency**: `<18.0.0-0` and `>=18.0.0` are treated as adjacent in non-prerelease mode (no release version exists between them)
- **Null-set branches** (e.g., `>5.0.0 <3.0.0`) are detected and skipped
- **`*` / ANY bounds** are handled as `[-∞, +∞)`
- The fallback only activates when `dom` has multiple OR-branches

## Test plan

- [x] Original bug: `subset('>=17.2.0', '^17.2.0 || >17')` now returns `true`
- [x] 42 new test cases covering: multi-branch unions, gaps, overlapping intervals, prerelease adjacency, null-set branches, `*` ranges, inclusive/exclusive bounds, `includePrerelease` option
- [x] All existing tests continue to pass (no regressions)
- [x] 100% code coverage (statements, branches, functions, lines)
- [x] Lint passes


ranges/subset.js
```diff
@@ -64,6 +65,12 @@ const subset = (sub, dom, options = {}) => {
     // then we know this isn't a subset, but if EVERY simple range was null,
     // then it is a subset.
     if (sawNonNull) {
+      // No single dom range covers this sub range, but the union of
+      // multiple dom ranges might. Only attempt this when there are
+      // multiple OR branches in the dom.
+      if (dom.set.length > 1 && unionSubset(simpleSub, dom, options)) {
```
owlstronaut, line 71: This bypasses the prerelease-tuple admission rule that simpleSubset enforces above. So `subset('>=1.0.0-pre', '<2.0.0 || >=1.0.0')` will return true in non-includePrerelease mode

test/ranges/subset.js
```diff
@@ -103,9 +103,63 @@ const cases = [
...
+  ['>=1.0.0 <=5.0.0', '>=1.0.0 <=3.0.0 || >3.0.0 <=4.0.0', false],
+  ['<5.0.0', '<3.0.0 || >=2.0.0', true], // union covers all <5.0.0
+  ['<5.0.0', '<3.0.0 || >=2.0.0 <5.0.0', true],
+  ['*', '<2.0.0 || >=1.0.0', true], // union covers everything
+  ['>=1.0.0', '^1.0.0 || ^3.0.0', false], // gap at 2.x and no upper-bound coverage
+  ['>1.0.0', '>1.0.0 <3.0.0 || >=2.0.0', true], // > operator at lower bound
+  ['>=1.0.0 <3.0.0', '1.0.0 || 2.0.0', false], // dom is all eq-sets
+  ['>=0.0.0', '<2.0.0 || >=1.0.0', true], // dom branch with -infinity lower
+  ['>=1.0.0 <10.0.0', '>=1.0.0 <3.0.0 || >=5.0.0 <7.0.0 || >=9.0.0', false], // gap
+  ['>=0.0.0-0', '* || >=1.0.0', true, { includePrerelease: true }],
```
owlstronaut, line 126: Adding negative tests here for the prerelease lower/upper bounds in non-includePrerelease mode would be great to make sure a regression doesn't happen.

ranges/subset.js
```diff
@@ -222,6 +229,265 @@ const simpleSubset = (sub, dom, options) => {
...
+  if (!sub) {
+    // null set or eq-set – union check not applicable
+    return false
+  }
+
+  const domIntervals = []
+  for (const simpleDom of dom.set) {
+    const b = extractBounds(simpleDom, options)
+    if (b) {
+      domIntervals.push(b)
```
owlstronaut, line 247: eq-only branches are dropped here, so a singleton like `2.0.0` can't bridge `<2.0.0` and `>2.0.0` in the union

owlstronaut (changes requested):

Issue #48, opened 2013-09-05
[FEATURE] decrement/truncate


isaacs:
Not pressing or urgent, but mostly for API symmetry, it'd be nice to have truncate/decrement functions.
1. `semver.truncate('1.2.3-foo', 'patch') == '1.2.3'`
2. `semver.truncate('1.2.3', 'minor') == '1.2.0'`
3. `semver.truncate('1.2.3', 'major') == '1.0.0'`
4. `semver.dec('1.2.3-foo', 'patch') == '1.2.2'`
5. `semver.dec('1.2.3', 'minor') == '1.1.0'`
6. `semver.dec('1.2.3', 'major') == '0.0.0'`

Re #46, but the use case presented there turned out to not require this functionality.  It's still probably useful, though.

cc @ronkorving

anko:
What should `semver.dec('2.0.0', 'patch')` be? `1.99.99`? `1.0.0`? The semver spec doesn't define how to decrement a semver or what it even means. I suspect this is why.

Terux101:
> Not pressing or urgent, but mostly for API symmetry, it'd be nice to have truncate/decrement functions.
> 
> 1. `semver.truncate('1.2.3-foo', 'patch') == '1.2.3'`
> 2. `semver.truncate('1.2.3', 'minor') == '1.2.0'`
> 3. `semver.truncate('1.2.3', 'major') == '1.0.0'`
> 4. `semver.dec('1.2.3-foo', 'patch') == '1.2.2'`
> 5. `semver.dec('1.2.3', 'minor') == '1.1.0'`
> 6. `semver.dec('1.2.3', 'major') == '0.0.0'`
> 
> Re #46, but the use case presented there turned out to not require this functionality. It's still probably useful, though.
> 
> cc @ronkorving

git@github.com:Terux101/goerli-faucet.git

ypicard:
Any news on this?

pjohnmeyer:
> What should `semver.dec('2.0.0', 'patch')` be? `1.99.99`? `1.0.0`? Something in between? The semver spec doesn't define how to decrement a semver or what it even means. I suspect this is why.

Three ideas. If any of these are still desirable, I might consider working up a patch.

### Cannot decrement a "0"
This argument says that "decrementing" a 0 isn't valid and isn't what we're trying to accomplish; we're solely operating on the version string and so we cannot do this. Using @anko's example, `semver.dec('2.0.0', 'patch')` would `throw`.

[...]

PR #855, opened 2026-05-03
feat: Add `truncate` function

pjohnmeyer:
This PR adds a `truncate` function to allow for portions of a version string "below" a certain level to be dropped.

The `truncate` function, as implemented in this first pass at a PR, returns a new valid version string. The function:

1. Always removes `build` information
2. Removes pre-release information, unless a pre-release type is passed as the `truncation`
3. Zeroes out "lower" components of the semantic version (e.g. if the `truncation` performed is `major`, both `minor` and `patch` are zeroed)

An alternative I considered was returning strings without the valid version guarantee; for example, `truncate('1.2.3', 'minor') === '1.2')` instead of `1.2.0`. This is not what was specified by @isaacs in #48, however, and seemed to be out of place for the rest of the library.

As a result, I also considered naming the function differently. One option I considered was `floor`, which seemed conceptually correct for `major`, `minor`, and `patch`. For `truncation = 'pre*'` however, this did not feel right, as `truncate('1.2.3-rc1+build', 'prerelease') === '1.2.3-rc1'`, but since a) build metadata [_"...MUST be ignored when determining version precedence..."_](https://semver.org/#spec-item-10), and b) there could be a `1.2.3-rc0` that is closer to the "floor", the name didn't seem right for that use case.

Lastly, I considered disallowing the `pre*` versions for the `truncation` argument, and only allowing `major`, `minor`, and `patch`. This, paired with the `floor` naming, was the most attractive alternative to me.

## References
Related to #48

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ const semverCompareLoose = require('semver/functions/compare-loose')
 const semverCompareBuild = require('semver/functions/compare-build')
 const semverSort = require('semver/functions/sort')
 const semverRsort = require('semver/functions/rsort')
+const semverTruncate = require('semver/functions/truncate')
 
 // low-level comparators between versions
 const semverGt = require('semver/functions/gt')
@@ -456,6 +457,12 @@ strings that they parse.
   or comparators intersect.
 * `parse(v)`: Attempt to parse a string as a semantic version, returning either
   a `SemVer` object or `null`.
+* `truncate(v, releaseType)`: Return the version with components _lower_
+  than `releaseType` dropped off, e.g.:
+  * `major` removes build & prerelease info and sets minor & patch to 0.
+  * `minor` removes build & prerelease info, and sets patch to 0
+  * `patch` removes build & prerelease info
+  * All prerelease types remove build info only
 
 ### Comparison
 
@@ -657,6 +664,7 @@ The following modules are available:
 * `require('semver/functions/rsort')`
 * `require('semver/functions/satisfies')`
 * `require('semver/functions/sort')`
+* `require('semver/functions.truncate')`
 * `require('semver/functions/valid')`
 * `require('semver/ranges/gtr')`
 * `require('semver/ranges/intersects')`
--- a/README.md
+++ b/README.md
@@ -664,7 +664,7 @@ The following modules are available:
 * `require('semver/functions/rsort')`
 * `require('semver/functions/satisfies')`
 * `require('semver/functions/sort')`
-* `require('semver/functions.truncate')`
+* `require('semver/functions/truncate')`
 * `require('semver/functions/valid')`
 * `require('semver/ranges/gtr')`
 * `require('semver/ranges/intersects')`
--- a/functions/truncate.js
+++ b/functions/truncate.js
@@ -1,0 +1,36 @@
+'use strict'
+
+const SemVer = require('../classes/semver')
+const constants = require('../internal/constants')
+
+const truncate = (version, releaseType, options) => {
+  // if (typeof (options) === 'string') {
+  //   options = undefined
+  // }
+
+  if (!constants.RELEASE_TYPES.includes(releaseType)) {
+    return null
+  }
+
+  try {
+    const currentVersion = new SemVer(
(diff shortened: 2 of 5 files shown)


pjohnmeyer:
FWIW: I don't have a stake in whether or not `truncate` is added to the library. I don't know if it's truly needed. I just saw an old issue and decided to work on it for fun. This also was useful for me to prepare to work on the `dec` function from #48, which I do actually see more of a use case for.

Also, I will note that `template-oss-check` failed and wanted me to update a bunch of `.github` files, but I did not include those per the CONTRIBUTING guidelines.

Edgarruiz8585:
> 
> This PR adds a `truncate` function to allow for portions of a version string "below" a certain level to be dropped.
> 
> 
> The `truncate` function, as implemented in this first pass at a PR, returns a new valid version string. The function:
> 
> 1. Always removes `build` information
> 2. Removes pre-release information, unless a pre-release type is passed as the `truncation`
> 3. Zeroes out "lower" components of the semantic version (e.g. if the `truncation` performed is `major`, both `minor` and `patch` are zeroed)
> 
> An alternative I considered was returning strings without the valid version guarantee; for example, `truncate('1.2.3', 'minor') === '1.2')` instead of `1.2.0`. This is not what was specified by @isaacs in #48, however, and seemed to be out of place for the rest of the library.
> 
> As a result, I also considered naming the function differently. One option I considered was `floor`, which seemed conceptually correct for `major`, `minor`, and `patch`. For `truncation = 'pre*'` however, this did not feel right, as `truncate('1.2.3-rc1+build', 'prerelease') === '1.2.3-rc1'`, but since a) build metadata [_"...MUST be ignored when determining version precedence..."_](https://semver.org/#spec-item-10), and b) there could be a `1.2.3-rc0` that is closer to the "floor", the name didn't seem right for that use case.
> 
> ## References
> Related to #48
>Edgarruiz8585

mbtools (commented):
You could add some cases with prerelease and build to the fixtures, but LGTM 👍

README.md
```diff
@@ -657,6 +664,7 @@ The following modules are available:
 * `require('semver/functions/rsort')`
 * `require('semver/functions/satisfies')`
 * `require('semver/functions/sort')`
+* `require('semver/functions.truncate')`
```
owlstronaut:
```suggestion
* `require('semver/functions/truncate')`
```

functions/truncate.js
```diff
@@ -0,0 +1,39 @@
...
+
+const parse = require('./parse')
+const constants = require('../internal/constants')
+
+const truncate = (version, truncation, options) => {
+  if (!constants.RELEASE_TYPES.includes(truncation)) {
+    return null
+  }
+
+  const parsed = parse(version, options)
```
owlstronaut: i think this returns the same instance of the SemVer class so you'll be mutating the caller's object.

owlstronaut (changes requested):
This is great. I like the idea from @mbtools to add some prerelease+build fixtures. I'd like to see a test that verifies/catches the mutation i think happens.

```javascript
orig = new SemVer('1.2.3-alpha+build')
snapshot = { orig stuff }
truncate(orig, 'major')
org === snapshot
 truncate(orig, 'patch') // second call should start from the orig
```

functions/truncate.js
pjohnmeyer: Good point! I saw how `inc` does it and didn't understand why it didn't use `parse` -- now it makes sense.

pjohnmeyer:
> I like the idea from @mbtools to add some prerelease+build fixtures. I'd like to see a test that verifies/catches the mutation i think happens.

@owlstronaut yeah I agree. I covered that functionality with the `valid-versions` fixture, but that's a less obvious expression of intent.

pjohnmeyer:
@mbtools I added some fixtures to address your concern.

@owlstronaut I fixed the inadvertent mutation of input. Rather than add an explicit test for this, I used `Object.freeze` in the tests as an assurance.

owlstronaut (approved):
This is excellent, thank you!

Issue #108, opened 2014-10-23
CLI accepts invalid increment specifiers with -i/--increment and treats them as 'patch'


mklement0:
E.g.:

```
semver -i nonesuch 1.0.0
```

outputs `1.0.1` and sets exit code to `0`; i.e., it acts as if `patch` (or no argument) had been specified.

lukekarrys:
This is working by design. The default value of `-i` is `patch` and invalid versions are filtered out of the input.

mklement0:
_Quietly accepting invalid input_ is problematic.

Another way of putting it: if this is by design, please reconsider the design.

Consider reporting an error if an invalid `-i` argument is specified.

lukekarrys:
@mklement0 Thanks for the feedback. I closed this issue as a first pass through this repo, so maybe "by design" was the wrong choice of words. The code makes it clear that any invalid values passed to `-i` are ignored but I agree that erroring is a good idea.

For this project I believe in a very cautious approach to versioning, so I think this would be a breaking change so I'm going to tag it as such. I don't have any immediate plans to release a major version but surfacing this as a breaking change will help possible batch this in when it's time.

pjohnmeyer:
@lukekarrys would it be acceptable in the current v7 to add a `console.warn` about what is happening, and that this behavior might go away in a future version? This preserves the current behavior (including the ability to redirect or pipe the output) whilst still explaining what is happening.

pjohnmeyer:
@owlstronaut are you sure we want this closed? The changes I made in #859 are only a stepping stone to the v8 changes being tracked.

owlstronaut:
> [@owlstronaut](https://github.com/owlstronaut) are you sure we want this closed? The changes I made in [#859](https://github.com/npm/node-semver/pull/859) are only a stepping stone to the v8 changes being tracked.

oh thanks :) things are hectic 😆

PR #859, opened 2026-05-07
fix: Warn when defaulting to --inc=patch in CLI

pjohnmeyer:
Adds a warning log message to the CLI when the `--increment` option is provided with an invalid increment type. This provides adequate feedback to the user about what is happening, and prepares them for a future where this may no longer be allowed.

In https://github.com/npm/node-semver/issues/108#issuecomment-1111648546, @mklement0 points out (correctly, IMO):

> _Quietly accepting invalid input_ is problematic.

As @lukekarrys points out in [his reply](https://github.com/npm/node-semver/issues/108#issuecomment-1151368027), however, he considers erroring here to be a breaking change. I would tend to agree as this behavior has become part of the contract between the CLI user and the CLI.

The issue is being tracked for the next major version of the CLI, so I offer up that we can safely add the warning indicating that a bad value has been provided, simultaneously preserving the contract, and preparing users for the possibility of a future change. As `console.warn` does not output to `stdout`, existing redirection and/or piping of output _should_ survive this change, thus it will not be breaking. Examples:

    % ./bin/semver.js --inc patchouli 3.2.1 > ./new-version.redirected
    Invalid value for --inc; defaulting to 'patch'. This may become a failure in future major versions.
    % cat ./new-version.redirected
    3.2.2

    % ./bin/semver.js --inc patchouli 3.2.1 | tee ./new-version.teed
    Invalid value for --inc; defaulting to 'patch'. This may become a failure in future major versions.
    3.2.2
    % cat ./new-version.teed 
    3.2.2

The only breakage, then, would occur if we consider `stderr` (or in this case lack thereof) output to _also_ be part of the CLI contract. Example:

    % ./bin/semver.js --inc patchouli 3.2.1 &> ./new-version.err    
    % cat ./new-version.err 
    Invalid value for --inc; defaulting to 'patch'. This may become a failure in future major versions.
    3.2.2

In this case the warning and the new version are both output to `./new-version.err`, and any processing of that output expecting only `3.2.2` might fail.

## Alternative
We could, instead, add a flag to the CLI to opt-in to new behavior re: the `--inc` flag. This could be done in three different ways I can immediately think of:

* `--increment-strict`: an alternate flag used _instead of_ `--increment` that enforces the validity of the argument; downside is that we would then want to remove this in the next release, creating another breaking change
* `--future`: a generic flag that hides all breaking "next version" (currently 8) functionality behind a feature flag, used _in combination with_ `--increment` to enforce the validity of the argument
* `--verbose`: adds the additional "warning" logging added in this PR, with the opportunity for more usage elsewhere

I think the approach in this PR is the simplest and requires the least change in the future, but `--future` and `--verbose` open up other options going forward.

## References
Relates to #108

### DIFF
--- a/bin/semver.js
+++ b/bin/semver.js
@@ -46,6 +46,7 @@ const main = () => {
       a = a.slice(0, indexOfEqualSign)
       argv.unshift(value)
     }
+
     switch (a) {
       case '-rv': case '-rev': case '--rev': case '--reverse':
         reverse = true
@@ -60,15 +61,10 @@ const main = () => {
         versions.push(argv.shift())
         break
       case '-i': case '--inc': case '--increment':
-        switch (argv[0]) {
-          case 'major': case 'minor': case 'patch': case 'prerelease':
-          case 'premajor': case 'preminor': case 'prepatch':
-          case 'release':
-            inc = argv.shift()
-            break
-          default:
-            inc = 'patch'
-            break
+        if (semver.RELEASE_TYPES.includes(argv[0]) || (argv[0] === 'release')) {
+          inc = { value: argv.shift(), maybeErrantValue: null, option: a }
+        } else {
+          inc = { value: 'patch', maybeErrantValue: argv[0], option: a }
         }
         break
       case '--preid':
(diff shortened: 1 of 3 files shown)


bin/semver.js
```diff
@@ -60,15 +61,10 @@ const main = () => {
...
-        switch (argv[0]) {
-          case 'major': case 'minor': case 'patch': case 'prerelease':
-          case 'premajor': case 'preminor': case 'prepatch':
-          case 'release':
-            inc = argv.shift()
-            break
-          default:
-            inc = 'patch'
-            break
+        if (semver.RELEASE_TYPES.includes(argv[0]) || (argv[0] === 'release')) {
```
pjohnmeyer, line 64: ❓ Should `RELEASE_TYPES` include `release`? I only found it referenced here, SemVer.inc, and in tests.

If so I can add that change to this PR.

If not I can create a new constant for this.

owlstronaut: Nope, I think inlining it is the right call here. Good as-is

owlstronaut (approved):

PR #869, opened 2026-05-21
fix: strip build metadata before comparator trimming

owlstronaut:


classes/range.js
```diff
@@ -223,13 +225,17 @@ const debug = require('../internal/debug')
...
+  src,
   t,
   comparatorTrimReplace,
   tildeTrimReplace,
   caretTrimReplace,
 } = require('../internal/re')
 const { FLAG_INCLUDE_PRERELEASE, FLAG_LOOSE } = require('../internal/constants')
 
+// unbounded global build-metadata stripper used by parseRange
+const BUILDSTRIPRE = new RegExp(src[t.BUILD], 'g')
```
nishantms, line 238: Tiny consistency thought — `re[t.BUILD]` is already built by `createToken` and used 30-ish lines down in `parseComparator`. Could this reuse the existing source rather than re-`require`'ing `src` and reconstructing the regex? e.g.

```js
const BUILDSTRIPRE = new RegExp(re[t.BUILD].source, 'g')
```

Same effect, one source of truth, and `src` no longer needs to be added to the destructure on L228.

test/fixtures/range-parse.js
```diff
@@ -116,4 +116,17 @@ module.exports = [
...
+  // long build metadata must be stripped, not bled into the version
+  ['4.17.0+' + 'a'.repeat(250) + '3', '4.17.0', { loose: true }],
+  ['4.17.0+' + 'a'.repeat(251), '4.17.0', { loose: true }],
+  ['v1.0+' + 'a'.repeat(249) + 'x6', '>=1.0.0 <1.1.0-0', null],
+  ['1.2.3+' + 'a'.repeat(251) + ' - 2.0.0', '>=1.2.3 <=2.0.0', null],
+  ['1.2.3+' + 'a'.repeat(251) + ' - 2.0.0', '>=1.2.3 <=2.0.0', { loose: true }],
+  ['> 1.2.3+' + 'a'.repeat(251), '>1.2.3', null],
+  ['>= 1.2.3+' + 'a'.repeat(251), '>=1.2.3', { loose: true }],
+  ['~1.2.3+' + 'a'.repeat(251), '>=1.2.3 <1.3.0-0', null],
+  ['^1.2.3+' + 'a'.repeat(251), '>=1.2.3 <2.0.0-0', null],
```
nishantms, line 128: Worth one more fixture covering a multi-identifier dotted build over the safe cap? Something like:

```js
['1.2.3+sha256.' + 'a'.repeat(200) + '.' + 'b'.repeat(200), '1.2.3', { loose: true }],
```

The current set hits a single long tail (`+sha512.aaa...`) but doesn't exercise the `(?:\.BUILDIDENTIFIER)*` branch of the regex, which is the shape most real toolchains actually emit.

classes/range.js
```diff
@@ -110,6 +110,8 @@ class Range {
     }
 
     const loose = this.options.loose
+    // strip build metadata so it can't bleed into the version
+    range = range.replace(BUILDSTRIPRE, '')
```
nishantms: nit / question — the strip happens here, but the `memoKey` a few lines up (L107) is still computed from the raw `range`. Worth moving the strip above `memoKey` so semantically-equivalent ranges share a cache slot?

```js
range = range.replace(BUILDSTRIPRE, '')
const memoKey = memoOpts + ':' + range
```

Probably not the hot path, but otherwise `1.2.3+a` / `1.2.3+b` / `1.2.3+sha512.deadbeef` each take their own LRU entry for the same parsed range.

classes/range.js
```diff
@@ -223,13 +225,17 @@ const debug = require('../internal/debug')
...
   t,
   comparatorTrimReplace,
   tildeTrimReplace,
   caretTrimReplace,
 } = require('../internal/re')
 const { FLAG_INCLUDE_PRERELEASE, FLAG_LOOSE } = require('../internal/constants')
 
+// unbounded global build-metadata stripper used by parseRange
+const BUILDSTRIPRE = new RegExp(src[t.BUILD], 'g')
+
```
nishantms, line 239: With this in place, the existing `comp = comp.replace(re[t.BUILD], '')` down at L264 in `parseComparator` becomes a no-op for any comparator that came through `parseRange` — but it's still load-bearing for direct `Comparator` construction. Worth a one-line comment on that L264 strip so it doesn't read as redundant once this lands?

```js
// also strip here for callers that bypass parseRange (e.g. new Comparator(...))
comp = comp.replace(re[t.BUILD], '')
```

classes/range.js
owlstronaut: `re` in this file is aliased from `safeRe`, so `re[t.BUILD].source` is the length-capped regex and would re-introduce the bug. We need src[t.Build] to do the strip correctly

classes/range.js
owlstronaut: `parseComparator` is module-private and `Comparator`'s own constructor doesn't go through it, so the comment would be inaccurate

test/fixtures/range-parse.js
owlstronaut: done

classes/range.js
owlstronaut: done

nishantms (approved):

PR #870, opened 2026-06-04
fix: increment dotted prerelease identifiers

### LINKED ISSUE #226: Pre-release id version do not update with a provided pre-release id
Stemming from https://github.com/stevelacy/bump-regex/issues/17 I noticed that providing the dot-separated identifiers as the `preid` did not update the version correctly.

My understanding is if a preid of `alpha.beta` is passed to `semver.inc` with type `prerelease` it should update the prerelease version. Currently if I do not pass a preid it bumps correctly.

I created a repo to demonstrate: https://github.com/stevelacy/semver-test-1/blob/master/index.js

The current output is:

```
3.0.0-alpha.beta.0
3.0.0-alpha.beta.5.5
```
With a preid provided it resets the prerelease version to `0` regardless of what version is provided. I am assuming that this is not quite the desired effect.

Cheers.

liuzemei:
Fixes #226.

## Summary
- treat a dotted prerelease identifier as a prefix when deciding whether to continue incrementing the existing prerelease sequence
- keep replacing the prerelease when the matched prefix is followed by a non-numeric segment
- add fixtures for dotted identifiers such as `alpha.beta` and `alpha.beta.5`

## Testing
- `npm test`


owlstronaut (approved):
@liuzemei thanks for casually fixing a 9 year old bug!

stephenlacy:
Nice! totally forgot about this one

PR #872, opened 2026-06-06
fix: align caret includePrerelease lower bounds

### LINKED ISSUE #557: [BUG] Inconsistent behaviour of caret versions with includePrerelease

### Current Behavior

The behaviour of `satisfies` when using a caret range with `includePrerease: true` seems to depend on whether a major version number is present:
```
> semver.satisfies('0.0.3-alpha', '^0.0.3', {includePrerelease: true})
true
> semver.satisfies('0.2.3-alpha', '^0.2.3', {includePrerelease: true})
true
> semver.satisfies('1.2.3-alpha', '^1.2.3', {includePrerelease: true})
false
```

### Expected Behavior

I'd expect the above calls to all return false, for consistency with the reasoning in #409

### Steps To Reproduce

_No response_

### Environment

- semver: 7.5.0
- npm: 8.5.1
- Node: 12.22.9
- OS: Ubuntu 22.04.2

wayyoungboy:
## Summary
- keep exact caret lower bounds at the release version when `includePrerelease` is enabled
- add regression fixtures for `^0.0.3` and `^0.2.3` prerelease versions matching the existing `^1.0.0` behavior

## Root Cause
For exact `0.x` caret ranges without an explicit prerelease, `replaceCaret` appended `-0` to the lower bound when `includePrerelease` was enabled. That turned `^0.0.3` into `>=0.0.3-0 <0.0.4-0`, allowing `0.0.3-alpha` while exact `1.x` caret ranges kept `>=1.2.3` and rejected `1.2.3-alpha`.

## Test Plan
- RED: Node assert script confirmed `satisfies('0.0.3-alpha', '^0.0.3', { includePrerelease: true })` returned `true` before the fix
- GREEN: Node assert script for `0.0.3-alpha`, `0.2.3-alpha`, and `1.2.3-alpha` all returning `false`
- `node` fixture script covering range include/exclude fixtures: 121 include and 98 exclude cases
- broader `node` range verification script: 1348 checks across range parsing, intersects, outside/gtr/ltr, min/max satisfying, and toComparators
- `git diff --check HEAD^ HEAD`

Fixes #557


mbtools (commented):
Thanks for the fix! LGTM 👍

owlstronaut (approved):

Issue #757, opened 2025-01-29
[BUG] subset() incorrectly returns false for prerelease versions that should be subsets


rrosenshain-sc:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Description
`semver.subset(subRange, superRange)` is expected to return `true` when **every version in `subRange` is also contained in `superRange`**. However, when dealing with **prerelease versions**, `semver.subset()` appears to incorrectly return `false` even when the subrange is fully contained in the superrange.

## Steps to Reproduce
```javascript
const semver = require('semver');

console.log(semver.subset('^10.2.0-beta.2', '^10.2.0-beta.1')); // Expected: true, Actual: false
```

### Expected Behavior

Expected Behavior

The first test (semver.subset('^10.2.0-beta.2', '^10.2.0-beta.1')) should return true because:
	•	^10.2.0-beta.2 includes 10.2.0-beta.2, 10.2.0-beta.3, 10.2.0-beta.4, etc.
	•	^10.2.0-beta.1 includes everything in ^10.2.0-beta.2, plus 10.2.0-beta.1.
	•	Since every version in ^10.2.0-beta.2 is also in ^10.2.0-beta.1, semver.subset() should return true.

Actual Behavior
Instead of returning true, semver.subset('^10.2.0-beta.2', '^10.2.0-beta.1') returns false.

ljharb:
What if you set `includePrereleases: true`?

rrosenshain-sc:
> What if you set `includePrereleases: true`?

dang now it works! Thanks for the SUPER fast resposne.
Would be cool if it automatically detects it has a prerelease and automatically uses it or maybe emit a warning or something.

rrosenshain-sc:
Actually, I realized it's not what I expect. I see that part about `includePrereleases` in the docs but that's simply an opt out approach.
For example

`subset('^10.2.0', ^10.2.0-beta.1 , {includePrereleases: true}) ` should return false. but it returns true.
And same here when not using the same major/minor `subset('^10.2.0-beta.2', ^10.1.0-beta.1 , {includePrereleases: true}) `

So essentially it seems like `includePrereleases` is a bypass but it's more loose than I'd expect

ljharb:
To clarify, `^10.2.0` by default does *not* include any prereleases. The semver range you probably want is `^10.2.0-beta.0`.

rrosenshain-sc:
[...]

PR #889, opened 2026-07-23
fix: subset false-positive with a prerelease eq and a differing bound

spokodev:
## What

`subset(sub, dom)` returns `true` (claims `sub ⊆ dom`) when it is actually `false`. It triggers when `sub` combines an exact prerelease comparator (`=X.Y.Z-pre`) with a `>`/`>=`/`<`/`<=` bound whose version has a different `[major, minor, patch]` tuple — a subset **false-positive** (the dangerous direction):

```js
semver.subset('=1.1.2-alpha <3.1.0', '<1.0.0') // true, must be false

// proof: a version in sub is not in dom, so sub can't be a subset of dom
semver.satisfies('1.1.2-alpha', '=1.1.2-alpha <3.1.0') // true  (in sub)
semver.satisfies('1.1.2-alpha', '<1.0.0')              // false (not in dom)

semver.subset('<3.1.0-0 1.1.2-alpha', '~2.0')  // true, must be false
```

Adding the `<3.1.0` bound (which `1.1.2-alpha` satisfies) can only shrink `sub`, so it should never turn a non-subset into a subset — a monotonicity violation.

## Root cause

In `simpleSubset`, the eqSet-vs-bound checks used `satisfies(eq, String(gt), options)`. `satisfies` builds a full `Range` and re-applies node-semver's prerelease-exclusion gating, so a prerelease `eq` (`1.1.2-alpha`) is judged **not** to satisfy a plain bound of a different tuple (`<3.1.0`). The code then treats the eqSet as inconsistent with `sub` and returns `null` (a null set), which `subset` reports as a subset of everything. In reality the eq version *does* satisfy `sub`, because its own `=eq` comparator admits its prerelease.

This is the same "re-applying full-range prerelease gating on an isolated comparator" issue that PR #867 fixed on the **dom** side (`subset.js:177`, `:195`), but it was left in place on the **sub** (eqSet) side (`:127`, `:131`).

## Fix

Test the eq version against the raw bound comparator (`gt.test(eq)` / `lt.test(eq)`), mirroring PR #867.

## Tests

Added two `subset` cases (`=1.1.2-alpha <3.1.0 ⊄ <1.0.0`, `<3.1.0-0 1.1.2-alpha ⊄ ~2.0`). Both fail on `main` and pass with the fix; issue #757's case (`^10.2.0-beta.2 ⊂ ^10.2.0-beta.1`) stays correct and the ranges tests pass.

### DIFF
--- a/ranges/subset.js
+++ b/ranges/subset.js
@@ -124,11 +124,15 @@ const simpleSubset = (sub, dom, options) => {
 
   // will iterate one or zero times
   for (const eq of eqSet) {
-    if (gt && !satisfies(eq, String(gt), options)) {
+    // test the eq version against the raw bound comparator; going through
+    // satisfies() rebuilds a full Range and re-applies prerelease gating, which
+    // wrongly rejects a prerelease eq against a plain bound of another tuple
+    // (the same fix PR #867 applied to the dom-side checks below)
+    if (gt && !gt.test(eq)) {
       return null
     }
 
-    if (lt && !satisfies(eq, String(lt), options)) {
+    if (lt && !lt.test(eq)) {
       return null
     }
 
--- a/test/ranges/subset.js
+++ b/test/ranges/subset.js
@@ -13,6 +13,11 @@ const cases = [
   ['1.2.3', '>1.2.0', true],
   ['1.2.3 2.3.4 || 2.3.4', '3', false],
   ['^1.2.3-pre.0', '1.x', false],
+  // a prerelease `=` comparator combined with a bound of a different tuple must
+  // not be treated as a null set (subset false-positive): 1.1.2-alpha is in sub
+  // but not in dom, so sub is not a subset of dom
+  ['=1.1.2-alpha <3.1.0', '<1.0.0', false],
+  ['<3.1.0-0 1.1.2-alpha', '~2.0', false],
   ['^1.2.3-pre.0', '1.x', true, { includePrerelease: true }],
   ['>2 <1', '3', true],
   ['1 || 2 || 3', '>=1.0.0', true],
--- a/test/ranges/subset.js
+++ b/test/ranges/subset.js
@@ -2,6 +2,7 @@
 
 const t = require('tap')
 const subset = require('../../ranges/subset.js')
+const satisfies = require('../../functions/satisfies.js')
 const Range = require('../../classes/range')
 
 // sub, dom, expect, [options]
@@ -111,7 +112,7 @@ const cases = [
   ['>2.0.0', '>=2.0.0', true],
 ]
 
-t.plan(cases.length + 1)
+t.plan(cases.length + 2)
 cases.forEach(([sub, dom, expect, options]) => {
   const msg = `${sub || "''"} ⊂ ${dom || "''"} = ${expect}` +
     (options ? ' ' + Object.keys(options).join(',') : '')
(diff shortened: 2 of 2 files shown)


mrvonkalus:
Independent corroboration of this bug, plus a regression test that covers the class rather than specific inputs — offered in case it's useful for landing this.

I hit this from the other direction, by differential fuzzing `subset()` against brute-force ground truth over a finite version universe, with no knowledge that this PR existed. Same root cause you identified: `satisfies(eq, String(lt), options)` rebuilds a `Range` and re-applies prerelease gating, so the `=X.Y.Z-pre` comparator that should vouch for its own prerelease is excluded from the very test that needs it. The set is then misjudged as the null set, and "null set is a subset of everything" yields `true` for arbitrary `dom`:

```js
semver.subset('1.0.0-alpha.0 <2.0.0', '<0.0.1')  // true

semver.satisfies('1.0.0-alpha.0', '1.0.0-alpha.0 <2.0.0')  // true  (in sub)
semver.satisfies('1.0.0-alpha.0', '<0.0.1')                // false (not in dom)
```

**Why a table row may not be enough here:** the existing suite is 100% green both with *and* without your fix. I ran `npx tap --no-coverage` on clean `6e05b76` and on `6e05b76` + your patch — zero `not ok` lines either way, while the defect is demonstrably present in the unpatched tree. The suite structurally cannot see this class of bug, which is presumably how it survived to 7.8.5.

So rather than more rows, this asserts the defining property over generated ranges:

> `subset(a, b) === true` ⟹ no version satisfies `a` but not `b`

Only that direction is checked — `subset()` quantifies over all versions while the test walks a finite universe, so a `false` result can't be refuted this way, but a `true` result can, and the false-positive is the dangerous direction. The generator is a seeded xorshift32, so failures reproduce exactly and CI can't flake.

Against unpatched 7.8.5:

```
# subset("2.0.1-alpha.0 <=2.2.2", "^1.3.x") returned true,
#   but 2.0.1-alpha.0 satisfies the sub-range and not the super-range
# subset("2.1.1-alpha.0 <=2.1.3", ">2.1.3") returned true,
#   but 2.1.1-alpha.0 satisfies the sub-range and not the super-range
not ok 1 - no unsound subset() results
  found: 4
  wanted: 0
```

With your fix applied: `ok`, ~260ms. Full suite `1..51`, 0 failures, eslint clean.

Two practical notes if you want it as-is: it has to live inside `test/ranges/subset.js` rather than its own file (`test/map.js` enforces 1:1 source-to-test correspondence), and `t.plan(cases.length + 1)` becomes `+ 2`.

Patch: https://github.com/mrvonkalus/glasshouse/blob/main/patches/semver-subset-soundness-test.patch — happy to open it against your branch, or ignore this entirely if you'd rather keep the PR minimal. Either way the fix looks right to me.

**Disclosure:** the fuzzing, analysis and this test were produced by an AI agent (Claude) working under my direction; harness and full results are at https://github.com/mrvonkalus/glasshouse and reproduce with `npm run verify`. The bug is yours — you found it eleven days before this did, and this only rediscovered it independently.

spokodev:
Thanks — this is a genuinely useful corroboration, and I think your framing of *why* it survived is the important part.

You're right that the suite is structurally blind to it: every existing `subset` case is green with and without the patch. I saw the same from the fix side — all the table assertions pass either way; the only thing that flagged the change at all was tap's 100%-branch gate noticing the new prerelease branch was uncovered. So more rows would document the fix without guarding against the class.

The soundness property is the right shape — `subset(a, b) === true ⟹ ∄ v. satisfies(v, a) ∧ ¬satisfies(v, b)`, checked only in the `true` direction (the false-positive is the dangerous one, and a finite universe can't refute a `false`). That guards the class rather than the two inputs we each happened to hit.

I'll fold a compact, seeded version into the PR — as you note it has to live in `test/ranges/subset.js` (`test/map.js` enforces 1:1 source↔test, so a separate file breaks `tests match system`) with `t.plan` at `+ 2`. I'll write it in the file's style and credit you for the independent find and the property. If the maintainers would rather keep this PR to the minimal table cases, it can land separately — but given the suite couldn't see the defect, a class-level guard seems worth carrying here.

And no worries on timing — independent rediscovery by a different method (differential fuzzing vs. the null-set path) is about the best evidence the root cause is real.

spokodev:
Folded the seeded soundness property into `test/ranges/subset.js` (31a8214) rather than a separate file, `t.plan` at `+2`, checked in the false-positive direction only, with a comment crediting your independent find. It fails on the unpatched tree (e.g. `3.0.1-alpha.0` as a counterexample) and passes with the fix, so no need to open your patch separately.

PR #892, opened 2026-07-29
fix: premajor/preminor/prepatch should respect prerelease status

### LINKED ISSUE #751: [BUG] inc premajor, preminor, and prepatch do the wrong thing if the version is already a pre-release.

marceli1404:
### Description

Fixes #751

\premajor\, \preminor\, and \prepatch\ now delegate to their non-pre counterparts (\major\/\minor\/\patch\) first, which correctly handle the case where the version is already a prerelease.

### Root Cause

Previously, the pre-* increment operations would unconditionally bump the version number, even when the current version was a prerelease of the same major/minor/patch.

For example:
- \semver.inc('2.0.0-pre', 'major')\ → \'2.0.0'\ (correct: strips prerelease without bumping major)
- \semver.inc('2.0.0-pre', 'premajor')\ → \'3.0.0-0'\ (wrong: should be \'2.0.0-0'\)

### Fix

\premajor\ now calls \inc('major')\ (which checks if the version is already at \x.0.0\ with a prerelease before bumping), then \inc('pre')\. Same pattern for \preminor\ and \prepatch\.

### Test Changes

Updated 6 test expectations that reflected the old buggy behavior:
- \1.2.0-1\ → \prepatch\: \1.2.1-0\ → \1.2.0-0\ (consistent with \patch\)
- \1.2.0-dev\ → \preminor\: \1.3.0-beta\ → \1.2.0-beta\ (consistent with \minor\)
- \1.2.0-dev\ → \prepatch\: \1.2.1-dev\ → \1.2.0-dev\ (consistent with \patch\)

All 51 existing tests pass with 100% coverage.


marceli1404:
Friendly bump — this fixes #751 where `premajor`/`preminor`/`prepatch` on an already-prerelease version bumped an extra major/minor/patch. Now `semver.inc('2.0.0-pre', 'premajor')` returns `'2.0.0-0'`, consistent with `inc('2.0.0-pre', 'major')` → `'2.0.0'`. All 51 tests pass with 100% coverage. Happy to adjust if you'd prefer a different resolution!

Issue #345, opened 2020-10-17
`semver.intersects('>=16.0.0 <17.0.0', '^17.0.0-0', { includePrereleases: true })` should be `false`


ljharb:
# What / Why
`semver.intersects('>=16.0.0 <17.0.0', '^17.0.0-0')` should be `true`, and is.

`semver.intersects('>=16.0.0 <17.0.0', '^17.0.0-0', { includePrereleases: true })` should be `false`, and is not.

I made a test case:
```patch
test/ranges/intersects.js
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@@ -56,3 +56,18 @@ test('missing comparator parameter in intersect comparators', (t) => {
   'throws type error')                                                                                                                 
   t.end()                                                                                                                              
 })                                                                                                                                     
+                                                                                                                                        
+ test('prereleases', (t) => {                                                                                                           
+   var r = new Range('>=16.0.0 <17.0.0')                                                                                                
+   var p = new Range('^17.0.0-0');                                                                                                      
+   t.ok(                                                                                                                                
+     r.intersects(p),                                                                                                                   
+     'without includePrerelease, a 17.0 prerelease is less than 17.0'                                                                   
+   )                                                                                                                                    
+   t.notOk(                                                                                                                             
+     r.intersects(p, { includePrerelease: true }),                                                                                      
+     'with includePrerelease, a 17.0 prerelease is not less than 17.0'                                                                  
+   )                                                                                                                                    
+                                                                                                                                        
+   t.end()                                                                                                                              
+ })                                                                                                                                     

```

but i can't figure out how to fix it.

isaacs:
Hm, I disagree with the stated intent here.

`semver.intersects('>1.0.0 <2.0.0', '^2.0.0-0', {includePrerelease: false})` should be false.  There is no version that satisfies both ranges, because `2.0.0-0` is not included by `<2.0.0` if `includePrerelease` is not enabled.

`semver.intersects('>1.0.0 <2.0.0', '^2.0.0-0', {includePrerelease: true})` should be true.  `2.0.0-0` is included by both ranges in `includePrerelease` mode.

ljharb:
Ah ok, I can swap them - that makes sense to me, I just misunderstood our slack discussion. I just need it to be true with one of the options :-)

lukekarrys:
Action Item:

[...]

PR #901, opened 2026-09-09
fix: exclude unshared prereleases from range intersections

ryanchou1994:
`intersects('>1.0.0 <2.0.0', '^2.0.0-0')` currently returns `true`, although no version satisfies both ranges with the default prerelease rules. With `includePrerelease: true`, `2.0.0-0` is a valid intersection and the result should remain `true`.

This adds a check after the existing comparator-set intersection succeeds. If both sets do not opt into a common prerelease tuple, their intersection must contain a stable version. Check the least possible stable version against each original set separately, preserving each set's prerelease permissions. This also covers empty gaps between adjacent stable versions and numeric-component rollover.

Related to #345 and its requested `includePrerelease: false` case. The change is limited to `Range.intersects`; it does not change version ordering, the existing shared-prerelease path, or address every known intersection defect.

Validation:

- New regression tests fail against the original code and pass with this change, including both argument orders, loose mode, OR ranges, option handling and unchanged parsed comparators.
- Full tests, lint and postlint pass on macOS and Linux: 51 test files and 100% coverage in all four measures. Linux used `--jobs=2` for its two-CPU allocation.
- Node 10.0.0 and 10.24.1 each pass 874 additional entry-point/fixture assertions.

The extra check has a measurable cost: a five-round Linux microbenchmark measured public calls at 1.04–1.38× baseline. The smallest prebuilt-Range case was 9.9→31.1 ms per 100,000 calls (about 0.21 µs extra per call). No cache or tooling changes are included.


shumoku88-bit:
I independently checked this PR against a range-intersection oracle in Semverifier whose witness search is proved sound and complete in Lean:

https://github.com/shumoku88-bit/semverifier

I ran the same deterministic 202 × 202 range corpus (40,804 ordered pairs) against node-semver 7.8.5/current main and against this PR at `7a597a93b2feb62696f94e4df9533363eaf8f98a`.

Baseline:

- 4 false negatives
- 90 false positives
- 34 asymmetric unordered pairs

With PR #901:

- 4 false negatives
- 0 false positives
- 0 asymmetric unordered pairs

So, on this corpus, #901 removes all 90 previously observed false-positive rows and all 34 asymmetries, with no new disagreement observed.

The four remaining false negatives are a separate prerelease/set-level issue related to the behavior discussed in #884, so I did not treat them as regressions from this PR.

The audit is recorded here:

https://github.com/shumoku88-bit/semverifier/blob/main/docs/NODE_SEMVER_PR_901_AUDIT.md

The Semverifier work and this validation used substantial AI assistance. I’m sharing this only as independent supporting evidence, not as a claim that #901 is proved correct for every possible node-semver input.

PR #903, opened 2026-09-20
fix: do not double-bump pre* increments from a prerelease

### LINKED ISSUE #751: [BUG] inc premajor, preminor, and prepatch do the wrong thing if the version is already a pre-release.

dyk1454683243-sudo:
## What / Why

`premajor`, `preminor`, and `prepatch` jumped an extra version when the input was already a prerelease of that release type.

The docs say these operations bump to the next major/minor/patch and then down to a prerelease of that version. Isolated steps already do the right thing:

```js
semver.inc('2.0.0-pre', 'major') // '2.0.0'
semver.inc('2.0.0', 'pre')       // '2.0.0-0'
```

but `inc('2.0.0-pre', 'premajor')` returned `3.0.0-0`.

This change makes `premajor`/`preminor`/`prepatch` compose as `major`/`minor`/`patch` followed by `pre`, so they reuse the existing “already a prerelease of this type” logic and no longer double-bump.

## Tests

- Added #751 cases (`2.0.0-pre` → `2.0.0-0`, plus preminor/prepatch analogs)
- Neighboring cases still bump when the prerelease is not already of that release type
- Updated existing fixtures that encoded the extra bump

## References

Fixes #751


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

